PR #22, opened 2013-09-30
Use jQuery events instead of addEventListener for IE8 support (when using excanvas)

TheJosh:
I haven't updated the minified version, and I don't expect you to pull this in it's current state because I'm guessing you don't want a dependency on jQuery, but nonetheless, here is a fix for anyone out there who wants to support IE8, and is using jQuery anyway.

### DIFF
--- a/signature_pad.js
+++ b/signature_pad.js
@@ -18,9 +18,10 @@
 var SignaturePad = (function (document) {
     "use strict";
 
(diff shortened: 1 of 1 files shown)


garygreen:
imo drop ie8 support altogether. Google has done this already. If you want IE8 support then include shims or fork.

szimek:
Closing it as there are no plans to support IE8.

Issue #48, opened 2014-03-10
Reject points generated by resting hand


efc:
Signature_pad does a great job of capturing finger-drawn signatures, but if a stylus is being used, it tends to capture a lot of artifacts generated by the user resting their hand on the glass while signing. This is particularly true on a larger device, like an iPad.

It would be great if there were some filter built in that would reject points too far away from the current curve to be legitimate signature points. An example of the problem is attached. Any thoughts on how to accomplish this?

![efc](https://f.cloud.github.com/assets/604967/2375403/1ebcab30-a86b-11e3-8254-a3b39b880595.png)

szimek:
Fixed by merging https://github.com/szimek/signature_pad/pull/57

PR #57, opened 2014-05-20
Feature/disable multi touch

jurreantonisse:
A proposed fix for #48 This disables multi touch by only using the first touchpoint.


szimek:
Thanks @jurreantonisse and sorry it took so long...

PR #72, opened 2014-10-07
Compatibility with AMD/CommonJS

jpsullivan:


bcherny:
+1 for grunt-umd. but there's no need to generate 2 versions, just generate 1, which will continue falling back to browser globals and being api-compatible with the current version.

szimek:
It uses `grunt-umd` now, so I'm closing this issue.

PR #85, opened 2014-12-16
Adds "Type to sign" support to Signature Pad.

cryptojuice:
We've added type to sign functionality to the signature pad.

Requires setting the following params:

``` javascript
options {
  enabledThypeToSign: true,
  inputSelector: "#selector-for-input",
  maxFontSize: "30",  // Defaults to 300 pt.
  fontface: "Arial" // Defaults to Helvetica
}
```


szimek:
I'm closing this one. It could be done as a separate library or a custom code (using [strokeText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/strokeText) method) that simply uses the same canvas as this library.

PR #90, opened 2015-02-11
Add destroy method that removes all the event listeners. 

miles-billsman:
This is so that if the signature pad is used in a SPA (for example in an Angular directive), the user can destroy it and prevent it from leaking memory.

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -24,6 +24,11 @@ var SignaturePad = (function (document) {
         this._handleTouchEvents();
     };
 
+    SignaturePad.prototype.destroy = function() {
+    	this._unhandleMouseEvents();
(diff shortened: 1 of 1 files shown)


szimek:
Thanks! One question - is there any reason to set all handlers to null after calling `removeEventListener`?

smcclstocks:
this would be really nice to have

szimek:
Closed by #106

hughanderson4:
I see in 2015 there was `destroy()` method added.

Some time between then and now it has been removed.  Why?

I put the Signature capture in a modal, now I need to properly remove any event handlers after modal is closed.

I need such `destroy()` method to prevent memory leak.

@szimek can we please add it back to the signature_pad?

Issue #93, opened 2015-03-09
Main entry missing from package.json


jamiecollinson:
This prevents `require('signature_pad")` from working with webpack / browserify. Suggested addition is:

```
"main": "./signature_pad.min.js",
```

szimek:
@jamiecollinson Could you make a PR?

szimek:
Closed by #96

PR #96, opened 2015-03-20
support browserify, addresses #93

chevett:
This is all that is needed for browserify support.  I've tested and it works great.


szimek:
Thanks!

PR #98, opened 2015-04-09
Relying on clientX and clientY is not safe.

webmaster777:
clientX and clientY may report incorrect values. It is better to use the page offsets. (pageX and pageY). This will fix issues with the signature pad being shown on a _scrolled_ and or zoomed page where the lines appear on a wrong location. I'm sure this happens on Android using e.g. Google Chrome. I left the previous code in. I'm not sure whether  pageX and pageY are widely supported.


szimek:
Could you prepare a jsfiddle with some instructions how to reproduce the incorrect behavior when scrolling/zooming the page?

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/pageX) pageX/Y are non-standard and according [UI Events spec](http://www.w3.org/TR/DOM-Level-3-Events/#h-interface-mouseevent), pageX/Y is supposed to be calculated based on clientX/Y.

richardcpeterson:
Sounds like maybe webmaster777 was running into the same bug I was with my pull request.  Is there a better solution to this than what webmaster or I suggest?

webmaster777:
Well, I was having a hard time trying to reproduce the issue in a small jsfiddle. Apparently, the location of the signature pad and probably the positioning of the surrounding elements had something to do with it.

PR #106, opened 2015-06-02
Add .off method that unbinds all event handlers

Rob-ot:


szimek:
Thanks! I'll check it out on Thursday and if everything is ok, I'll release 1.4.0 with this patch.

Rob-ot:
Thanks!

szimek:
Thanks! Though, I suspect a new issue to re-enable signature pad to be reported really soon :)

PR #115, opened 2015-09-04
Add an .on() method to re-bind event

Alplob:
As guessed in #106 , I think that adding an .on() method to re-bind events after the method .off() was called would be useful (at least it is in my current project ^_^).


szimek:
@Alplob Thanks! Couldn't it just call
```javascript
this._handleMouseEvents();
this._handleTouchEvents();
```
though?

Alplob:
Yes, you are right, that is also ok!

PR #123, opened 2015-09-29
Prevent duplicate events on tap in iOS Safari

### LINKED ISSUE #120: when trying to dot the proverbial i the dots almost always result in two dots
it seems the library always tries to create two points when trying to make a dot... 

tempfileforshare_2015-09-21-11-27-53

Device: Samsung Galaxy Note 4 (5.1.1)

PerfectPixel:
This change prevents iOS Safari from firing an mouse down & mouse up event in addition to the touch start and touch end event if you just tap the screen.

This resolves #120.

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -58,6 +58,7 @@ var SignaturePad = (function (document) {
         this._handleTouchEnd = function (event) {
             var wasCanvasTouched = event.target === self._canvas;
             if (wasCanvasTouched) {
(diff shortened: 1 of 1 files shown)


szimek:
I'll try to check it out on some Android and iOS devices over the weekend and merge it if everything is fine.

szimek:
Thanks!

PerfectPixel:
You are very welcome!

PR #130, opened 2015-10-20
Restrict form submission

itsazzad:
If the signature stays in a form then clicking on the button was submitting the form. So now its prevented.


szimek:
Thanks! But in this case I think it's just easier to just add `type="button"` attribute to button elements. By default it's set to `"submit"` on [button elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type) and that's why clicking it submits a form, if you put it inside one.

itsazzad:
:+1:

szimek:
I fixed using the `type` attribute.

PR #150, opened 2016-02-18
Fix touchend event on touch devices

mtomic:
It fixes touchend event on hybrid apps such as Cordova. It now fires ontouchend event correctly and saves the data from the Canvas.


szimek:
I merged just 7021319fa13eb5395d352f5dc21864272fda4f91, as the rest of files is generated automatically. Thanks!

mtomic:
:+1:

PR #165, opened 2016-06-09
Fix issue with smaller image on mobile device

olegabr:
If use c.fromDataURL(c.toDataURL()) from one canvas to another on the same page, it draws smaller image.
It can be reproduced on both, desktop browser with a scale changed, and mobile browser.

The idea of this fix is to eliminate ratio from calculations. It seems that modern browsers do it themselves.


UziTech:
fixed in #253

olegabr:
Wow! just 5 years later! I'm impressed :-)

UziTech:
Actually it was fixed in 2017 I'm just closing old issues and prs

PR #166, opened 2016-06-27
Fix focus

Teemu:
I noticed this problem on mobile devices where if you have input selected and you have pressed back button to hide, the keyboard reappears after you start signing.

At the beginning where the input has the focus but user has hidden the keyboard with back button:
screenshot_20160627-120704 1

After trying to sign something, the keyboard pops up:
screenshot_20160627-120714

This isn't a perfect solution. Stealing focus is not ideal **1)** if you have external keyboard and there might be a case where **2)** user wants to have the keyboard up AND sign at the same time. However, that is not very common and the focus stealing already happens on desktop.

### DIFF
--- a/example/css/signature-pad.css
+++ b/example/css/signature-pad.css
@@ -86,6 +86,13 @@ body {
   }
 
 .m-signature-pad--footer
+  .description
+    input {
+      margin-left: 10px;
+      width: 60px;
+}
+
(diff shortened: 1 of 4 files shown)


Teemu:
I'm going to close this PR since 1) it's been so long and probably not going to get merged anytime 2) I no longer work at the same company.

Issue #74, opened 2014-10-23
Store signature data as data:image/svg+xml;base64


PSchreuder:
I love this simple signature library! Thanks! There's one thing I really want to implement, and that's storing the signature data as svg / data:image/svg+xml;base64. So I can store the full-vector signature in my SQL database, and use it for high quality PDF's/print.

What's the best way to retrieve the signature in data:image/svg+xml;base64?

yhurynovich:
Hi, I would like to sponsor SVG functionality (svg export). Can you please estimate cost of building of this function for me?

BioPhoton:
I also need SVG export! 
@szimek are you interested in building this functionality?

szimek:
I don't think I'll have enough time to do it. It shouldn't require that many changes though: 
1. store all points in memory, instead of just the 4 last ones
2. generate SVG from all these points using native Bézier curves (i.e. `<path d="Mx,y C...">`); I have no idea how one can draw Bézier curves with variable width in SVG :/

If anyone wants to prepare a PR, it would be awesome.

BioPhoton:
Hi,
i guess an easy solution for implementing this feature would be to parse the canvas after drawing with a canvas to svg lib.

I did some quick tests with following libs:
- https://github.com/gliffy/canvas2svg
- https://github.com/kangax/fabric.js

With the canvas2svg lib i got propper output but the lines are not visible anymore

I replaced 
https://github.com/szimek/signature_pad/blob/master/signature_pad.js#L53
with
 this._ctx = new C2S(500,500);

With fabric i got also svg output but i guess i use it in a wrong way.

@szimek do you think this would be a good solotuin without much work.

szimek:
It might be, but it would be awesome if somebody actually created a working prototype :) I'm afraid that the way I'm drawing lines (using filled circles) might not work very well with SVG unless some post-processing is used to smooth lines out.

If it doesn't work properly, it might be easier to prepare SVG by hand, but only if one assumes that all curves have the same width and just use the [C syntax](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d#Curveto).

jackspirou:
@szimek I submitted a PR for saving as svg: https://github.com/szimek/signature_pad/pull/173

feedback welcome

mhbu50:
1+

BioPhoton:
THX to @jackspirou for this great pr!

oliverjanik:
merge?

[...]

PR #173, opened 2016-07-27
add save as svg

jackspirou:
relevant to issue #74


Nic128:
@jackspirou It works well, except if you try using `fromDataURL` with what you exported with `toDataUrl('svg')`. It ignores the viewBox from the svg, and stretches the stroke.

I found another tiny bug. If you export to png, then reapply the png to the canvas using `fromDataURL`, then try to export to svg, the svg is empty.

mhbu50:
+1

jackspirou:
TL;DR: I'm not sure that functionality is needed by the `signature_pad` library, and some of it may be very difficult to pull off. 

@Nic128 unfortunately the intention of this pull request was only to save the signature as an `svg` not read an `svg` formatted file, or try to convert from `svg` -> `png` or `png` -> `svg`. To be honest i'm surprised that the `fromDataURL` method works taking an `svg` data string at all.  The `svg` image is computed during the drawing/execution of the user's signature in real time.  The easiest way to accurately use that real time computed information (such as velocity width) is when the line is drawn on the canvas. When trying to go back after the line is drawn, it is harder to determine what the velocity width is. Infact most canvas to `svg` libraries I found seem to create an `svg`/canvas object that is used in lieu of the normal canvas `ctx`. 

It might be possible to iterate over every `x, y` coordinate within the canvas, read that value, and if it is not white/blank then append it to an `svg` element, but I would be dubious if that method would result in decent renderings to `svg`. As for reading `svg` you could do something like http://jsfiddle.net/Na6X5/ and write the svg to the canvas. Then you could use the iteration method described above to recreate the `svg` output. 

I hope that explanation satisfies why you saw the behavior described in the `fromDataURL` and `toDataUrl` methods. I think this library is focused on rendering signatures from the HTML5 canvas element into `svg` or `png` format, not necessarily reading `svg` onto the canvas, converting `svg` -> `png` or vice versa.  If you have a good argument for those features mentioned above I would love to hear it, and/or  @szimek's thoughts on the matter.

Wheph! @Nic128 all that said, you are welcome to add that functionality onto this PR as a PR of your own :)  

Thanks!

szimek:
@jackspirou Hey, sorry for taking so long to review it, but I was busy and somehow missed it completely. Thanks a lot for this PR. I'll try to review it today.

2 issues though after skimming through the code:
- it would be great to put some of this code (e.g. for setting svg up, clearing it, etc.) into separate functions, so the existing functions don't get much longer
- if the library kept all the points, generated curves and whatever is needed in memory (instead of keeping just the last 4 points), wouldn't it be possible to calculate all this SVG stuff only when `toDataURL('image/svg')` is called? this way it wouldn't affect the performance while drawing and maybe the `svgOn` wouldn't be necessary?

jackspirou:
Hi @szimek, no worries.  I've got no particular timeline and we all get busy :) 

> it would be great to put some of this code (e.g. for setting svg up, clearing it, etc.) into separate functions, so the existing functions don't get much longer

I agree from a code extensibility point of view.  Different formats then could have their own "cleanup" methods and such.  It would be a trivial change if you plan to support more formats. 

 > if the library kept all the points, generated curves and whatever is needed in memory (instead of keeping just the last 4 points), wouldn't it be possible to calculate all this SVG stuff only when toDataURL('image/svg') is called? this way it wouldn't affect the performance while drawing and maybe the svgOn wouldn't be necessary?

Yes it is possible if you kept all the points and associated velocity width in an array of "point objects" you could iterate through.  I think it would be a major change to the library if the PNG rendering used this array as well.  My hope was to add SVG as a minor design addition, not touch existing PNG functionality.  I think its a valid design (and would remove `svgOn`), just not one that was in scope for the original PR. 

From a performance perspective I haven't seen any issues rendering the SVG in realtime since the svg object is in memory anyway.  I would be interested to know if you have had any performance issues in the past with something like this.  I assumed the 4 point buffer was there to keep memory consumption low, but maybe moving from 4 points to all points with their associated velocity width is acceptable.  You might have done more general testing on devices and know the answer to that question?

I like the idea of having this data in memory, and different rendering functions then simply convert that array to SVG or PNG or X.  It seems like a very extensible, clean design pattern.  Again it would be a larger design change and I think thats probably your call.

szimek:
> Yes it is possible if you kept all the points and associated velocity width in an array of "point objects" you could iterate through. I think it would be a major change to the library if the PNG rendering used this array as well. My hope was to add SVG as a minor design addition, not touch existing PNG functionality. I think its a valid design (and would remove svgOn), just not one that was in scope for the original PR.

The PNG rendering would keep working as it is right now, after all JPEG/PNG encoding of canvas contents is handled natively by the browser. The drawing on the canvas would also keep working as it is, you need to see what you draw as you draw it 😄 I'm thinking about keeping those points just for SVG for now, though maybe the library could expose them, so that you could send them to your server and do some strange processing there. The drawback is that SVG rendering will be visually slightly different than what you see on the canvas.

Another thing is that keeping all those points might help to finally write some basic automated tests 😉 However, to do that the library would need to accept these points as input as well...

> From a performance perspective I haven't seen any issues rendering the SVG in realtime since the svg object is in memory anyway.

The issue is that on older mobile devices (especially Android phones) there's a small noticeable lag as you draw - it's tiny, but rather annoying and I'd like to prevent adding any code that might increase it. I haven't done any testing though... I'd love to do some performance profiling to decrease this lag, but really don't have time how to do it.

jackspirou:
> PEG/PNG encoding of canvas contents is handled natively by the browser.

Ahhh.  Yeah that makes sense.  Yeah, i overlooked that somehow :) 

> The issue is that on older mobile devices (especially Android phones) there's a small noticeable lag as you draw - it's tiny, but rather annoying and I'd like to prevent adding any code that might increase it. I haven't done any testing though... I'd love to do some performance profiling to decrease this lag, but really don't have time now to do it.

I got my hands on a 2011 Motorola Razor with the original android browser and the performance seemed fine, but I can see how methods attached to the SVG object such as `setAttributeNS` and `appendChild` could be taxing.  `appendChild` can be slow and I found a somewhat similar comment about `setAttributeNS` from w3.org: https://lists.w3.org/Archives/Public/www-dom/2011OctDec/0223.html.

szimek:
I don't know much about SVG, so maybe let's do it like this. I'll keep this PR opened for now. It looks like you don't actually need points, but rather curve objects and width for each curve, so I'll update the code (most likely `_addPoint` function), so that it stores all curves and their widths in some variable and maybe you could then modify `toDataURL` method to generate SVG using the code from this PR. Once the new PR works and is merged, I'll close this one.

What do you think?

jackspirou:
@szimek sounds good, but to clarify you mean the `curve` object used here: 

```
var attr = "M "+curve.startPoint.x.toFixed(3)+","+curve.startPoint.y.toFixed(3)+" ";
    attr += "C "+curve.control1.x.toFixed(3)+","+curve.control1.y.toFixed(3)+" ";
    attr += curve.control2.x.toFixed(3)+","+curve.control2.y.toFixed(3)+" ";
    attr += curve.endPoint.x.toFixed(3)+","+curve.endPoint.y.toFixed(3);
```

I think some representation of points beyond curves are still needed for cases where someone "dots an i" and such, unless I misunderstand.

```
var c = document.createElementNS("http://www.w3.org/2000/svg", "circle");
    var dotSize = typeof(this.dotSize) === 'function' ? this.dotSize() : this.dotSize;
    c.setAttributeNS(null, "r", dotSize);
    c.setAttributeNS(null, "cx", point.x);
    c.setAttributeNS(null, "cy", point.y);
    c.setAttributeNS(null, "fill", this.penColor);
    this._svg.appendChild(c);
```

szimek:
@jackspirou I've finally found some time to work on it. There's a new branch [data-to-url-svg](https://github.com/szimek/signature_pad/tree/data-to-url-svg) that stores everything in `this._data` variable - it's an array of objects with `type` and `data` properties that store curves and dots (hopefully with all necessary widths). Could you check if it's enough to generate the SVG? If it is, could you modify `toDataURL` method on the `data-to-url-svg` branch, so that it will call a separate function that generates SVG if the `imageType` is `image/svg` or the existing code otherwise?

szimek:
@jackspirou @meuru-muthuthanthri @Nic128 @mhbu50 Hey, I've just pushed a new version to [draw-signature-from-point-data](https://github.com/szimek/signature_pad/tree/draw-signature-from-point-data) branch (you can find built files in `dist` folder now) that supports generating SVG files when you pass `'image/svg+xml'` to `toDataURL` method. It would be awesome if you could check it out and see if there are any errors.

I've modified code a lot (switched to ES6 with Babel, did some refactoring etc.), so it's possible that there are some new bugs there as well.

jackspirou:
Thanks @szimek for seeing this through.  I'll close out this PR in favor of yours and take a look at the  draw-signature-from-point-data branch. 👍

szimek:
@jackspirou I've updated the demo on this branch and added "Save as SVG" button - it opens a new window with base64 encoded SVG image. I've actually used a lot of your code there.

mymattcarroll:
@szimek, just curious why you moved the build files to a `dist` directory. Now, anyone that was requiring in the file directly using `./node_modules/signature_pad/signature_pad.js` (I believe RequireJS needs to be done this way) will have to make changes to their code.

szimek:
@mymattcarroll No specific reason, just to clean up stuff in the root folder. However, in `package.json` file there's `main` field that points to `dist/signature_pad.js` file, so it should be enough to just write `import SignaturePad from 'signature_pad'` or `require('signature_pad')`. Though I'd have to actually verify that it really works.

mymattcarroll:
If only everyone was using the same module loading system... (I wish)

We are currently have and older site using [grunt-contrib-copy](https://www.npmjs.com/package/grunt-contrib-copy) to copy all required files to a directory before concatenation and minification which requires a path to the the relative paths to required files.

Im sure there are other examples where people are using a path to get the file. Im happy to change our code if/when we update to new version. Might be worth clearly noting in the `CHANGELOG` that the path has changed though.

szimek:
@mymattcarroll Good point - I'll add info about it when I bump the version and update the `CHANGELOG`. Unless someone finds new bugs in this version, I'll release it as a beta version over the weekend.

szimek:
I've just release 1.6.0-beta.1 with this feature. This version is already used by https://szimek.github.io/signature_pad. Let me know if you find any issues.

chitgoks:
i tried to use 1.6.0-beta.1 and .toDataURL('image/svg+xml') returns 

data:image/png;base64,iVBORw0KGgoAAAANSUh ....

just wish to clarify, if svg+xml is the desired output, the contents should be those of svg tags, right? or it's still a base64? please advise.

szimek:
@chitgoks Are you sure you're using 1.6.0-beta.1? I've just tried it on the [demo page](http://szimek.github.io/signature_pad/) and got the following:
```js
signaturePad.toDataURL('image/svg+xml')

"data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zd…JsYWNrIiBmaWxsPSJub25lIiBzdHJva2UtbGluZWNhcD0icm91bmQiPjwvcGF0aD48L3N2Zz4="
```

chitgoks:
hi szimek. apologies it seems i placed another comment in another thread and deleted that instead of this.

it is working using your beta. great plugin!

szimek:
@chitgoks no problem. cheers!

chitgoks:
hi @szimek there were posts regarding trimming of canvas and while their solutions worked, it was with png. in this case, i would like to trim the svg.

i thought of getting the instance of signature pad's canvas, clone it, trim it, then create a new instance of signaturepad then calling toDataURL('...svg') but that did not work out.

would there be a possibility to add a trimming when saved to SVG?

PR #175, opened 2016-08-08
Add touchAction to allow scrolling when signature pad is off

felixhammerl:
This commit allows scrolling via touch when the pad is switched `.off()`. This was not possible before, since `touch-action` was set to `none` before.


nolybom:
Thanks for the optimization. One question though...

When a mobile user tries to scroll, the handleTouchStart event is immediately triggered. Do you know of any method we could kind of like delay that when user scrolls up/down on the element for the first time...or a better approach: if the canvas is visible for a period of time? What do you think?

this._handleTouchStart = function (event) {
    if (event.targetTouches.length === 1) {
      var touch = event.changedTouches[0];
      self._strokeBegin(touch);
    }
  };

szimek:
@felixhammerl It took some time, but it should be fixed in the next release ;) Thanks!

PR #193, opened 2016-10-20
Update signature_pad.js

josephfischer:
Added API onAddPoint so that calling applications can get an array of points instead of an image.  Most retail applications collect an array of points instead of an image.


szimek:
I'm closing this one, because in 1.6.0 (currently beta) one can get raw data using `toData()` method.

PR #208, opened 2017-01-11
Fixed isEmpty returning incorrect value after svg generation

mymattcarroll:
- Fixed `signaturePad.isEmpty()` returning `true` after calling `signaturePad.toDataURL('image/svg+xml')`

- Fixed generated svg containing NaN values when drawing curves with sharp corners

- Some simple refactoring to extract common code into a private function


src/signature_pad.js
```diff
@@ -426,31 +413,56 @@ SignaturePad.prototype._toSVG = function () {
...
+  svg.setAttributeNS(null, 'width', canvas.width);
+  svg.setAttributeNS(null, 'height', canvas.height);
 
-      // Draw a dot
+  this._fromData(
+    pointGroups,
+    (curve, widths) => {
+      const path = document.createElementNS('http;//www.w3.org/2000/svg', 'path');
+
+      if (!isNaN(curve.control1.x) &&
```
szimek: Could you add a comment why these checks are needed? I think I noticed NaN value in generated SVG once or twice, but I'm still not sure why they sometimes appear.

PR #209, opened 2017-01-12
Crop option, drawImage function and resize function

mymattcarroll:
### Added

* `signaturePad.crop` option to allow for `signaturePad.toDataURL()` to crop signature of whitespace before generating a dataURL. Creates a temporary canvas element to avoid modifying the canvas displayed.
* `signaturePad.drawImage(canvasImageSource)` function to allow for a [CanvasImageSource](https://developer.mozilla.org/en-US/docs/Web/API/CanvasImageSource) to be drawn onto the signaturePad. Will be scaled down to fit the canvas if the image does not fit and will also centre the image on the canvas.
* `signaturePad.resize(width, heigh, scaleDown)` function to resize the canvas without clearing the signature.

### Tested On:

* Chrome: 55.0.2883.95
* Firefox: 49.0.2 and 50.1.0
* Safari: 10.0.2
* Opera: 42.0
* Android: 4.4.4 kitkat (using [Cordova](https://cordova.apache.org/docs/en/latest/guide/platforms/android/)) on a Samsung Galaxy J1


szimek:
This is the best looking and well tested PR so far :) 

One thing though - I'm really not sure I want to have all this stuff in the library, as it will make maintaining it more cumbersome. All of these methods could be rewritten as utility functions that have nothing to do with signature pad library itself, but simply manipulate a canvas element - whatever is drawn on it already.

The undo feature you mentioned before is specific to this library and while I have no idea if it would be useful to anyone (or at least a significant number of people, which would justify adding it and all the maintenance cost that goes with it), it might eventually be added if there are enough people who want it.

I'll think about it.

mymattcarroll:
@szimek, could you please close this PR if you still do not wish to include these features in the code base? Ill need to use these features in a production setting shortly and if will need to create a new library that includes this functionality. Thanks for the consideration and keep up the awesome work!

mymattcarroll:
Closing after changing base branch to master for PR in #218

PR #216, opened 2017-01-30
Reduce Resulting SVG; Trunacte Trailing 0s; Use Styles

pmn4:
@szimek cannot thank you enough for this repo. Here's a suggestion for reducing the resulting SVG by about half. It replaces inline styles with a `style` element, uses relative data points instead of absolute, and truncates insignificant zeros from coordinates.

signature-svg

_I don't have babel running on my machine, so wasn't able to produce the dist versions. if you like this change, feel free to copy the changes and close the PR. thanks again!_

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -436,6 +436,11 @@ SignaturePad.prototype._toSVG = function () {
   const maxY = canvas.height;
   const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
 
+  const style = document.createElementNS('http://www.w3.org/2000/svg', 'style');
+  // TODO: is there a proper API for creating styles?
+  style.innerHTML = `/* <![CDATA[ */ path {stroke:${this.penColor};fill:none;stroke-linecap:round; /* ]]> */`;
+  svg.appendChild(style);
+
   svg.setAttributeNS(null, 'width', canvas.width);
   svg.setAttributeNS(null, 'height', canvas.height);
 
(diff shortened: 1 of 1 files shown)


src/signature_pad.js
```diff
@@ -451,16 +456,20 @@ SignaturePad.prototype._toSVG = function () {
...
+        const diff2x = curve.control2.x - curve.startPoint.x;
+        const diff2y = curve.control2.y - curve.startPoint.y;
+        const diffEx = curve.endPoint.x - curve.startPoint.x;
+        const diffEy = curve.endPoint.y - curve.startPoint.y;
+
         const attr = `M ${curve.startPoint.x.toFixed(3)},${curve.startPoint.y.toFixed(3)} `
-                   + `C ${curve.control1.x.toFixed(3)},${curve.control1.y.toFixed(3)} `
-                   + `${curve.control2.x.toFixed(3)},${curve.control2.y.toFixed(3)} `
-                   + `${curve.endPoint.x.toFixed(3)},${curve.endPoint.y.toFixed(3)}`;
+                   + `c ${diff1x.toFixed(3)},${diff1y.toFixed(3)} `
```
pmn4: likely lost in the shuffle is the subtle change from `C` (absolute coordinates) to `c` (relative coordinates)

szimek:
@pmn4 Thanks! I'll try to review it this week and, if all is ok, merge it and release as a second beta. 

Your PR made me notice one thing though - the data structure that is used to draw SVG and is returned by `toData` method stores only info about points. It doesn't store pen color or width settings for each line. If someone draws e.g. each line using a different pen color, exported SVG won't have this info and all lines will have the same color... It's probably a very rare case, but I'm not sure now if this info should be stored and if code that generates SVG should take it into account...

Adding new features sucks 😉

pmn4:
ah, right! I saw `this.penColor` and starting thinking about the wrong `this`!
Well, take what you like and discard the rest.
Thanks again for this repo

UziTech:
closing as stale

PR #223, opened 2017-02-13
Add bower setup instructions

guumo:
Add bower setup instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,10 @@ or Yarn:
 ```bash
 yarn add signature_pad
 ```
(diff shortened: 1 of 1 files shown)


szimek:
@guumo I actually wanted to stop publishing it as a bower package, because I wanted to get rid of built files from the repo and, if I understand it correctly, I'd need to have them there for bower.

szimek:
@guumo Thank you for the PR, but Bower is no longer supported. The last supported version is 1.5.3.

PR #237, opened 2017-03-30
throttling-feature added

kunukn:
> Hey, now that 1.6.0 is released, could you prepare a PR with your code from http://codepen.io/kunukn/full/bgjzpb/? I'd merge it and release as 1.7.0. I guess the throttle function could be simplified a bit, because we won't be using leading/trailing options.

To test this. You use this in the **app.js**
```javascript
signaturePad = new SignaturePad(canvas, {
  throttle: 16 // x milli seconds
});
```

The throttle.js is 34 lines and is from Underscore. Well-tested and production ready.

I tried this 22 lines example which could be used instead.
https://remysharp.com/2010/07/21/throttling-function-calls

I tried this 12 lines example, but could not get it to work correctly with my Codepen example. Not sure why.
http://sampsonblog.com/749/simple-throttle-function

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -1,5 +1,6 @@
 import Point from './point';
 import Bezier from './bezier';
+import throttle from './throttle';
 
 function SignaturePad(canvas, options) {
   const self = this;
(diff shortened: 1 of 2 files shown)


szimek:
@kunukn Big thanks for this feature, especially that it took so long to add... I've just released 2.1.0-beta.1 with it and if nobody finds any issues, I'll release it as 2.1.0 soon. 

BTW. I set throttling to 16 by default.

PR #241, opened 2017-04-04
don't stretch image in fromDataURL

sadiqkhoja:
don't stretch the image if it is smaller than canvas and scale down proportionately if it is larger.

problem

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -94,14 +94,31 @@ SignaturePad.prototype.clear = function () {
 
 SignaturePad.prototype.fromDataURL = function (dataUrl) {
   const image = new Image();
-  const ratio = window.devicePixelRatio || 1;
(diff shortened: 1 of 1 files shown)


src/signature_pad.js
```diff
@@ -94,14 +94,31 @@ SignaturePad.prototype.clear = function () {
...
+  const deviceRatio = window.devicePixelRatio || 1;
+  const width = this._canvas.width / deviceRatio;
+  const height = this._canvas.height / deviceRatio;
 
   this._reset();
   image.src = dataUrl;
   image.onload = () => {
-    this._ctx.drawImage(image, 0, 0, width, height);
+    const imgWidth = image.width / deviceRatio;
+    const imgHeight = image.height / deviceRatio;
```
MartijnR: This code works for me if imgWidth and imgHeight are not divided by deviceRatio.

Issue #200, opened 2016-12-07
Image size on mobile device


zazics:
"fromDataURL()" incorrectly resize image data on Android device.
Image data is saved correctly into a local file using "toDataURL()" but when I reload it, image is half its original size.
This is due to pixel ratio. I made a quick fix in file "signature_pad.js" in "fromDataURL" function line 124.
I remove the divide by pixel ratio and now image size is correct:
width = this._canvas.width /*/ ratio*/,
height = this._canvas.height /*/ ratio*/;

kashban:
I would rather recommend to enhance signaturePad to keep aspect ratio of a given image and fit it into a canvas of a given size.

lexon0011:
I have the same problem. Is there already a solution for this problem?

UziTech:
fixed in #253

Issue #153, opened 2016-03-10
Bug in fromDataURL for high DPI screen devices


siggifv:
# Description

I've found that the method `fromDataURL` behaves oddly on devices that don't have the pixel ratio 1, e.g. on mobile devices with high DPI.
I'm aware of the [clause in the README](https://github.com/szimek/signature_pad/blob/master/README.md#handling-high-dpi-screens) where handling of high DPI screens is discussed. 
However, I'm still able to reproduce an issue using `fromDataURL` in cases where the canvas is of a fixed size. I believe that [this scaling of the image](https://github.com/szimek/signature_pad/blob/master/signature_pad.js#L124-L125) is the reason for this issue.
You can reproduce this issue using [this JSFiddle](https://jsfiddle.net/siggifv/1oLghs5w/).
## Steps to reproduce
1. Open the JSFiddled linked in the description above on a device that **does not have a pixel ratio of 1** and write something on the signature pad.
2. Click Save.
   ![1_after_save](https://cloud.githubusercontent.com/assets/1277384/13671375/85dcdadc-e6c7-11e5-932d-195f76b09a3b.PNG)
3. Click Edit.
4. Write some more on the pad.
   ![2_before_cancel](https://cloud.githubusercontent.com/assets/1277384/13671378/8f9fe186-e6c7-11e5-884f-7d4bb2997182.PNG)
5. Click Cancel
### Expected outcome

The signature pad should look exactly like it did after step 2.
### Actual outcome

The changes on the signature pad are reverted, but the image has been scaled so it looks different from what was on the pad after step 2.
![3_after_cancel](https://cloud.githubusercontent.com/assets/1277384/13671384/9407cd24-e6c7-11e5-94de-1925664bc8fa.PNG)

durasj:
Hi,

I also stumbled upon this issue. This is unexpected behavior. I think we should have a choice whether we want such modifications (so second argument defaulting to previous behavior should be fine). Mainly because something like that is not done automatically anywhere else. 

Thanks a lot @siggifv for your investigation, you saved me some time.

Manually populating canvas is the way for now:

``` js
                    var image = new Image();
                    signaturePad.clear();
                    image.src = dataURL;
                    image.onload = function () {
                        signatureCanvas = document.querySelector("#signature-pad");
                        signatureCanvas.getContext("2d").drawImage(image, 0, 0, width, height);
                    };
                    this.signaturePad._isEmpty = false;
```

RonB:
[...]

Issue #105, opened 2015-05-15
FromDataURL reducing image size


rmmackay:
Hello,

If I save a signature to a database using the toDataURL() function, and then I load that string back into the canvas using the fromDataURL() function, then the resulting image appears in the canvas but is noticeably smaller than the original image. 

To test, draw in the canvas along all 4 edges. Save and then reload. You will see that the right-side and bottom-sides of the image now have space between the image and the border because the image was made slightly smaller.

Since I re-save images over and over for a good reason, they keep shrinking with each save!!!

Help!

razvanphp:
Hmm, this happens only on Retina Display, I just reproduced it. You must deal with it yourself somehow, the creator said already multiple times that the library should not take care of this.

razvanphp:
It seems that this solves the problem:
http://jsfiddle.net/4JH75/19/

Mottoweb:
I am experiencing the same issue with device pixel ratio 1. Any help?

durasj:
Related to https://github.com/szimek/signature_pad/issues/153

demo-igor:
To make the same image size you can add options
`signaturePad.fromDataURL(data, {width: 100, height: 100})`
width, height should be the same as canvas size
More details:
https://github.com/szimek/signature_pad/blob/master/src/signature_pad.ts

iDevIt007:
Thanks to your last answer, we can now fix the problem with fromDataURL scaling issues, should close this issue.

pavankumar0425:
Thanks a lot for the response "demo-igor"

Issue #89, opened 2015-02-07
fromDataUrl fails on tablet


mathpere:
Hi,

I use this awesome lib on my current project and I meet some issue with tablet.

First of all, all is working fine on desktop browser (mac non-retina, chrome / firefox).

The problem occured on tablet (nexus 7, android lollipop, Chrome) with fromDataURL: The signature size is shrank on tablet.

Here is a demo: http://jsbin.com/tixuje/2/edit?html,js,output

If you open this link: http://jsbin.com/tixuje/2 on desktop, you will see this one:
![Original signature capture](https://docs.google.com/uc?id=0B_ksY-Q-HtWOQlJsckNqWk9UVXc)

When I open this same link: http://jsbin.com/tixuje/2 on my tablet, I get a shrank signature:
![Shrank signature capture](https://docs.google.com/uc?id=0B_ksY-Q-HtWOOVVYQnYycEZQQ0E)

Do you know where does the problem?

mathpere:
After checking, if this can help, the problem doesn't occur on iPad (safari + chrome)

szimek:
I don't own a tablet with Android, but hopefully I'll be able check it on my friend's Nexus 5 on Monday. Does it also happen if you draw the image on Nexus 7 and then read it back?

mathpere:
I also own a nexus 5: it happens too with it.

szimek:
@mathpere I still haven't tried it out on Nexus 5, but I prepared some tests, tried it out on non-retina and retina Macbooks and it seems to work fine.

Here's the first jsfiddle with image created on non-retina Macbook (`devicePixelRatio = 1`): http://jsfiddle.net/szimek/tge7vy68. It looks like it's correctly displayed on both Macs.

Here's the first jsfiddle with image created on retina Macbook (`devicePixelRatio = 2`): http://jsfiddle.net/szimek/4utubzhp/. Again, it's correctly displayed on both Macs. One thing to notice here, is that __the image is actually 4x bigger than the one created on non-retina device__, but it's still displayed correctly.

I used this jsfiddle to create those images: http://jsfiddle.net/szimek/u4zmsmbL. If you click "Get data" button, it will open a new window where you can copy the data URL from.

This HiDPI stuff is pretty confusing, so it's still possible that my tests are somehow messed up and it doesn't work correctly after all.

Let me know if it actually helps in any way :)

mathpere:
Hi,

Thanks for your support.

I don't know yet how to interpret these tests.

[...]

PR #253, opened 2017-05-16
Allow custom ratio/width/height in fromDataURL

halo:
Dear @szimek,

I know you're as busy as I am :) I'd still kindly ask you consider this PR as a small change that potentially solves a number of issues and PRs in one strike.

It simply allows a custom ratio, width, and height to be passed in when loading an image onto the canvas.

It should solve these Issues:

* The last comment in #89 suggests a monkey patch that is solved by this PR
* #105 appears to be solvable by overriding width/height via `fromDataURL` as implemented in this PR
* #153 complains about guessing the ratio automagically, should be solved by this PR
* #200 describes just the problem that this PR solves

And you should be able to close these PRs:

* #241 suggests to shrink any larger image to the canvas. While this makes sense at first sight, it's probably magic that may not fit all. This PR allows the developer to do the calculation outside of the SignaturePad library and simply pass in the resulting width and height.
* #165 butchers away any ratio guessing. That only solves one specific problem. This PR solves it by passing in `ratio: 1` as an argument.
* #164 removes the ratio altogeher. That also works for that person's use case but is not desireable for others. This PR solves it by passing in `ratio: 1` as an argument.

If there is anything more I could do to assist, let me know.

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -92,12 +92,11 @@ SignaturePad.prototype.clear = function () {
   this._isEmpty = true;
 };
 
-SignaturePad.prototype.fromDataURL = function (dataUrl, options) {
+SignaturePad.prototype.fromDataURL = function (dataUrl, options = {}) {
   const image = new Image();
-  const opts = options || {};
-  const ratio = opts.ratio || window.devicePixelRatio || 1;
-  const width = opts.width || (this._canvas.width / ratio);
-  const height = opts.height || (this._canvas.height / ratio);
+  const ratio = options.ratio || window.devicePixelRatio || 1;
+  const width = options.width || (this._canvas.width / ratio);
+  const height = options.height || (this._canvas.height / ratio);
 
   this._reset();
   image.src = dataUrl;
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -92,11 +92,12 @@ SignaturePad.prototype.clear = function () {
   this._isEmpty = true;
 };
 
-SignaturePad.prototype.fromDataURL = function (dataUrl) {
+SignaturePad.prototype.fromDataURL = function (dataUrl, options) {
   const image = new Image();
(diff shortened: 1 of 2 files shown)


mymattcarroll:
👍

halo:
Yes, no, maybe? It would be sad if I had to uphold a fork. It's not a complex change. But if you don't have time in the next weeks, just let me know. So I won't have to follow up. Thank you!

(I'm a just little bit suspicious when I see issues and pull requests that are several years old :)

szimek:
Hey, huge thanks for this PR! In general I think it's a good idea, but I simply haven't got time yet to check it out. Hopefully, I'll find some time during this weekend.

halo:
Sure, just wanted to ask. No stress. I'm grateful you created this library in the first place! :D

src/signature_pad.js
```diff
@@ -92,11 +92,12 @@ SignaturePad.prototype.clear = function () {
   this._isEmpty = true;
 };
 
-SignaturePad.prototype.fromDataURL = function (dataUrl) {
+SignaturePad.prototype.fromDataURL = function (dataUrl, options) {
```
szimek: After rewrite to ES6, you can probably set the default values for function parameters, so you could write it as 
```js
function (dataUrl, options = {}) {...}
```
and get rid of `opts` variable.

halo: Indeed. I wasn't sure so I followed the pattern of other functions.

I changed it now, this is what it compiles to:

```js
SignaturePad.prototype.fromDataURL = function (dataUrl) {
  var _this = this;

  var options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};

  var image = new Image();
  var ratio = options.ratio || window.devicePixelRatio || 1;
  var width = options.width || this._canvas.width / ratio;
  var height = options.height || this._canvas.height / ratio;
  // ...

```

PR #260, opened 2017-06-07
Export to SVG with correct pen colors

DynamoEffects:
Hey @szimek this is a simple modification to export to SVG with the correct pen color for each stroke.

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -160,6 +160,7 @@ SignaturePad.prototype._strokeUpdate = function (event) {
     x: point.x,
     y: point.y,
     time: point.time,
+    color: this.penColor,
   });
 };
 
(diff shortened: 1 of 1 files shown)


src/signature_pad.js
```diff
@@ -404,7 +406,7 @@ SignaturePad.prototype._toSVG = function () {
 
         path.setAttribute('d', attr);
         path.setAttribute('stroke-width', (widths.end * 2.25).toFixed(3));
-        path.setAttribute('stroke', this.penColor);
+        path.setAttribute('stroke', color);
```
szimek: If there's no default value for color here (and for drawing dots as well), will it work with the existing exported SVG strings? Won't it set color to `undefined`?

DynamoEffects: Let's test and see...

DynamoEffects: So this works:

1. Create a design with multiple colors.
2. Export to SVG saves design with correct color combinations.
3. Import from the same SVG shows color combinations correctly.

What doesn't work:

At that point you cannot export to SVG again using 2.1.1 even without the proposed change since the SVG export feature relies on recording the strokes to the this._data array.

The only solution I see would be to repopulate the this._data array by reinterpreting the imported SVG data instead of relying on the canvas itself to draw the imported data.

szimek: Thanks! I've added a note in the README that calling `#toData` won't really work if signature was loaded using `#fromDataURL`.

PR #261, opened 2017-06-26
Fix bug when style.zoom is used

martinsdan:
Previously, setting a zoom in the document (document.body.style.zoom = x)
would create a scale multiplier that SignaturePad did not account for. This caused the
line to show up offset from where the user attempted to draw.

With this commit the zoom is taken into account when creating a point.

The bug can be seen here https://jsfiddle.net/0veL8mmb/

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -146,8 +146,8 @@ SignaturePad.prototype._strokeBegin = function (event) {
 };
 
 SignaturePad.prototype._strokeUpdate = function (event) {
-  const x = event.clientX;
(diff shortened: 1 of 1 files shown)


szimek:
@martinsdan Thank you for this PR, but according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/zoom) `zoom` property is non-standard and shouldn't be used on production sites. I don't really want to add workarounds for non-standard features.

UziTech:
closing as stale

Issue #245, opened 2017-04-14
Signature not getting captured on Microsoft surface pro 4


ghost:
Hi,

I am using a stylus to enter my signature using app developed on Microsoft surface pro 4. I am not able to capture the signature from the same. It works if I don't use a stylus.

Current behavior:
User enters signature using stylus. Signature gets entered. However the base64 of the signature image is not obtained.

Expected behavior:

User enters signature using stylus. Signature gets entered. However the base64 of the signature image is obtained.

Version of signature pad : Latest
Target device : MS Surface pro 4

szimek:
Hi @karanbalkar I don't understand. Does it mean that when using a stylus nothing gets drawn or that `#toDataURL` doesn't return anything?

Ben855:
Starting 7/3/2017 the stylus stopped working (nothing gets drawn on screen).  It had worked previously.
Using finger it will display.
The devices are MS Surface pro 4
I upgraded to the latest version of signature pad without effect.

szimek:
@Ben855 Which browser are you using? Edge? Which version exactly? Looks like when you use stylus it uses some other events than mouse and touch events...

Ben855:
I am using Chrome browser

Ben855:
Did further testing It will draw in IE.
In Chrome, using the Demo it will work initially, 
      but then stops after doing clear, have to refresh to get it to write again. 
      I do have two canvases defined on the page.

szimek:
Unfortunately, I don't own MS Surface and it's not available on Browserstack, so it's hard for me to debug what's actually going on there...

Ben855:
I need to borrow 1 to be able to debug.  
Hopefully I can debug it in the next few days.

Ben855:
Using the Demo, it is calling mouse events the first time I draw.  After I remove the stylus from the canvas it does not receive mouse events anymore from the stylus ( I can't draw anything else with the stylus

farshadsh:
Same issue here, it was working before but now it does not, it's working with FireFox, Edge, even Google Canary but Google Chrome which is wired.

onigunn:
I can confirm this. Using the demo: Initially you are able to draw. After you move stylus from the screen and try to continue drawing nothing happens.

Chrome: 59.0.307
Surface 4 Pro with Stylus

szimek:
[...]

PR #281, opened 2017-08-10
added support for PointerEvents

onigunn:
In order to capture events from input devices like a microsoft stylus,
we have to handle pointer events too.

Fixes issue #245

### DIFF
--- a/example/js/signature_pad.js
+++ b/example/js/signature_pad.js
@@ -114,6 +114,7 @@ function SignaturePad(canvas, options) {
   var self = this;
   var opts = options || {};
 
+  this.supportsPointerEvents = window.PointerEvent !== 'undefined';
   this.velocityFilterWeight = opts.velocityFilterWeight || 0.7;
   this.minWidth = opts.minWidth || 0.5;
   this.maxWidth = opts.maxWidth || 2.5;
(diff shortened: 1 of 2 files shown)


szimek:
@onigunn Thanks! I'll most likely release some other stuff first and then will release a version with your PR as a beta release. 

One question though: Chrome on desktop supports pointer events as well, won't it trigger event handlers twice - once via mouse events and once via pointer events?

onigunn:
@szimek I first thought so but Firefox behaves strange on PointerEvent detection. Mouse events stopped working. So I left the condition out. After some digging around it seems that pointer events must call `event.preventDefault()` to keep mouse events going. See: [MDN PointerEvent Reference](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent)

I changed that but currently have to wait for the surface and stylus to check if everything working as expected there too. I'll keep you updated and will update my PR next week.

szimek:
@onigunn Thanks! Unfortunately I don't have access to a tablet with stylus, so I'm not sure if there's any way for me to actually test it. I'll wait for your update then ;)

onigunn:
@szimek Sorry for the delay. Test device was missing :) 

Seems to work. Tested on Edge and Chrome on given Surface Pro 5 (or 4 - thb I don't exactly know which version) with Stylus. Updated my PR.

szimek:
@onigunn Thanks! I'll release a beta version of 2.4.0 with this PR in the next few days and see if it causes any new issues. If it doesn't, I'll release it as stable.

szimek:
Hey, I finally had some time to look into it and, unfortunately, I'm not really confident about merging it without testing it first. I'll try to finally find some device with a stylus ;)

You wrote:
> I first thought so but Firefox behaves strange on PointerEvent detection.

Most likely it is caused by broken detection code:
```js
window.PointerEvent !== 'undefined'
``` 
should either be 
```js
window.PointerEvent !== undefined
``` 
or 
```js
typeof window.PointerEvent !== 'undefined'
```

onigunn:
@szimek 
Yeah, totally right. Detection is broken here. Sadly I cannot provide you anymore changes because I don't have test device anymore :)

If you like you can close this PR. Or maybe a test device will show up and it will be merged.

szimek:
@onigunn Thank you again for your PR! I've just added an initial support for pointer events in the latest beta version (3.0.0-beta.3).

PR #300, opened 2017-09-13
Fix `minDistance` not being correctly initialized when set to zero.

### LINKED ISSUE #299: minDistance option can't be initialized to 0
Reporting a bug
The new option to set `minDistance: 0` during the signature pad initialization does not work, since `0` is skipped in the `0 || 5` evaluation. [JavaScript OR Evaluation](https://stackoverflow.com/a/2100787/1611947)

In the following example the option is set to `0`, and the resulting minDistance is displayed below the signature pad. I expected to see it assign `0`, but instead it assigned `5`.
https://jsfiddle.net/bvf29kdn/1/

Using v2.3.0. The assignment of 0 bug is on the following line:
https://github.com/szimek/signature_pad/blob/v2.3.0/src/signature_pad.js#L13

The following could potentially solve this issue: (Similar to how "throttle" is assigned).

```
this.minDistance = 'minDistance' in opts ? opts.minDistance : 5;
```

Thanks! (I can submit a pull request to fix as indicated above if that sounds good, just let me know!)

remomueller:
Closes #299.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### master
+#### Bug Fixes
+*   Fix `minDistance` not being correctly initialized when set to zero. ([remomueller](https://github.com/remomueller) closes [#299](https://github.com/szimek/signature_pad/issues/299).
+
 ### 2.3.0
 #### Bug Fixes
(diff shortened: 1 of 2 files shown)


szimek:
Awesome, thanks! I guess I finally need to start writing tests at some point...

Issue #309, opened 2017-10-03
The same bug , fromData and toData , the draw is not completely the same .


tempbing:
@szimek 
Tonight , I want to take signature_pad in my another project ( it must support "undo")... 
I'm sure that , this following bug has always been there ( signature_pad v2.3.2 ). So I submit this issue .

The same bug , fromData and toData , the draw is not completely the same .
https://github.com/szimek/signature_pad/issues/270
 
Do you have the impression ?

> 
> Thanks very much for reply.
> On this fixed version: (https://jsfiddle.net/szimek/h80phxyr/11/)
> I see the problem again .
> You can draw some curves (like six letters "jjJJjj") ,then click "undo" button six times .
> On this online demo.: (https://jsfiddle.net/szimek/h80phxyr/11/)
> I have a try ,maybe i don't understand ...
> The picture compare.
> http://huobei.oss-cn-shenzhen.aliyuncs.com/02.png
> http://huobei.oss-cn-shenzhen.aliyuncs.com/01.png
> ( Can you see my picture ?
> Other pic link:
> http://foreign-oss.oss-us-west-1.aliyuncs.com/02.png
> http://foreign-oss.oss-us-west-1.aliyuncs.com/01.png
> )

szimek:
Thanks for submitting it. I won't have time to look into it in the next 2-3 weeks, so it would be great if anyone could take a look at this issue and come up with a fix ;)

PieterVanPoyer:
Hello,

I think your problem can be solved by setting following properties:
```
throttle: false,
minDistance: 1
```
http://jsfiddle.net/PieterVanPoyer/7q0hzzuy/

The default properties makes the signature_pad more performant and smoother, but they makes the undo action a little more unpredictable.
Defaults:
```
throttle: 16,
minDistance: 5
```
I've created the fiddle with these 2 properties. Please, try your testcase again.
I'd love to submit a fix for this issue later. I could handle the strokeUpdate event differently, but I 'd have to find a way to make it work mobile (without the _mouseButtonDown flag).

UziTech:
closing as stale

PR #312, opened 2017-10-28
The same bug , fromData and toData , the draw is not completely the s…

PieterVanPoyer:
…ame . #309

When drawing jjJJjj like in the bug report and the throttle function waits 16ms by default.
So, if you do a mouseUp when the throttle timeout is running, the last 15 ms could be appended to the next undo section. If you draw quick.
I've flushed the throttle function on mouseup or when touching ends.
The throttle function behaves kind of strange. F.e. when you set the throttle to 1000ms it feels strange for me. But that is another issue.


PieterVanPoyer:
The lastPointIsNeverTooClose functionality is not absolutely necessary. So I gonna submit another pull request without the lastPointIsNeverTooClose functionality.

PR #313, opened 2017-10-30
https://github.com/szimek/signature_pad/issues/309 Flush the throttled function as solution

PieterVanPoyer:
When drawing jjJJjj like in the bug report and the throttle function waits 16ms by default.
So, if you do a mouseUp when the throttle timeout is running, the last 15 ms could be appended to the next undo section. If you draw quick.
I've flushed the throttle function on mouseup or when touching ends.

see: https://github.com/szimek/signature_pad/issues/309

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -1,6 +1,6 @@
 import Point from './point';
 import Bezier from './bezier';
-import throttle from './throttle';
+import Throttler from './throttle';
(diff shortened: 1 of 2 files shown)


szimek:
It should be fixed in the latest release 3.0.0-beta.2. If it's still broken, please reopen this PR.

Issue #249, opened 2017-05-03
it's possible to draw outside the box, when mouse is not released


laDanz:
**Do you want to request a feature or report a bug?**
BUG

**What is the current behavior?**
On a mobile device (I use ASUS VivoTab Note 8 (Windows 10 Home Version 1511) with Chrome 58.0.3029.96) you can draw outside the box if you begin the signature within the canvas (using your finger), and then leave the canvas(while still holding the finger down) and eventually re-enter the canvas and moveup the finger.
SignaturePad will collect data making the resulting points/svg/... exceed the given bounds.

Please see the screenshot, made with this fiddle: https://jsfiddle.net/t2kbbr4j/
![reproduction](https://cloud.githubusercontent.com/assets/166216/25659449/1ec7eaae-3008-11e7-8d75-b981b48acfa2.png)

**What is the expected behavior?**
I would expect that the SignaturePad behaves like in the non-mobile version, where you cannot draw outside the box.

**Which versions of SignaturePad, and which browser / device are affected by this issue? Did this work in previous versions of SignaturePad?**
Tablet: ASUS VivoTab Note 8 (Windows 10 Home Version 1511) 
Browser: Chrome 58.0.3029.96
SignaturePad: v2.1.1
Did this work in previous versions of SignaturePad? I don't know.

szimek:
Uh, I'll try it out on Android tablet (I don't have a Windows one) and see if I can replicate. If it really does return points outside of the canvas, I guess I'll have to check if coordinates are inside a canvas (i.e. are not negative and not higher than canvas dimensions)...

The previous versions of signature-pad (prior to 2.0.0) didn't return point coordinates, so it could be broken before, but wasn't visible.

arnoldmanzano:
Confirmed seeing this issue. Observed in Android 6/7 and iOS safari.
Able to re-create it by using Chrome dev-tools Device toolbar emulating a mobile-touchscreen device.

szimek:
I don't have that much time work to on this library, so it would be great if someone could create a PR with a fix for this issue. 

Maybe it would be enough to simply check if point coordinates are outside of the canvas (i.e. smaller than 0 or greater than height/width of the canvas), or maybe it would result in some strange curves being drawn...

arnoldmanzano:
Hi szimek, created a PR for this #324. Your suggestion seems to work well.

UziTech:
closing as stale

SamuelWillis:
[...]

PR #324, opened 2017-12-07
Fix for #249  - points are always inside the canvas

arnoldmanzano:
Fix for #249 - points are always inside the canvas


UziTech:
closing as stale

PR #327, opened 2017-12-28
Pad size initialization

### LINKED ISSUE #326: Size initialization failure when initial size of canvas (width and height) are specified on CSS
This jsfiddle show-up the actual issue
https://jsfiddle.net/gianluigiliguori/h0cg3s8u/
I expect that using CSS instead of HTML Attributes for define the canvas size the signature-pad would be fine.
I am using the last version of signature-pad on Chrome v63.0.3239.84

liguori:
Ensure correct signature pad initialization if canvas size attributes are not provided but its style is changed (ex. via CSS) 
Closes #326


UziTech:
closing as stale

PR #341, opened 2018-02-20
Added callback parameter to fromDataURL

### LINKED ISSUE #340: Pass callback to .fromDataURL()
Since formDataURL loads an image asynchronously, it would be great if it could return a Promise. I don't think it will cause any backwards compatibility issues.

I'd be happy to create a PR if interested.

MartijnR:
closes #340

Thanks for this cool library!

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -96,7 +96,7 @@ SignaturePad.prototype.clear = function () {
   this._isEmpty = true;
 };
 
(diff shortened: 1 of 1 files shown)


szimek:
This is implement in the latest release 3.0.0-beta.2. I needed it to write a test for `fromDataURL` and completely forgot about this PR :/

PR #347, opened 2018-03-07
Detect and handle closed shape events

costinEEST:
Trigger an event when the line being drawn crosses an existing line on the same path. 
Also, trigger an event whenever another line is crossed first.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,9 +6,11 @@ module.exports = {
   },
 
   "rules": {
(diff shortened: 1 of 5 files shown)


leye0:
This is a major feature when designing a drawer allowing shapes, altought this is more useful for drawer where you can tap to add vertices. Any opinion or idea about this feature request?

UziTech:
closing as stale

PR #351, opened 2018-03-12
set backgroundImage

costinEEST:
Add backgroundImage to options argument when SignaturePad is instantiated.


UziTech:
closing as stale

PR #353, opened 2018-03-27
Add onAddPoint() configuration callback

juanghurtado:
Adds a `onAddPoint()` callback to configuration options. This is executed on every time a point is added.

It may be a little bit overkill, but it depends on what you do inside the callback. I find it useful, it would be cool to have it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ signaturePad.on();
 <dd>(function) Callback when stroke begin.</dd>
 <dt>onEnd</dt>
 <dd>(function) Callback when stroke end.</dd>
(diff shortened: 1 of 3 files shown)


juanghurtado:
Any updates about this merge? Do you need me to do something else?

Thanks!

UziTech:
@szimek There are a lot of these PRs for emitting events maybe we should make signaturePad an event emitter to make adding these callbacks easier?

UziTech:
Ya I don't think it is a good idea to add a lot of `onAbc` options but I think the event emitter for the next version would be a good idea to make signature_pad more extensible.

PR #369, opened 2018-06-13
Add current maxWidth in IPointGroup for correct undo drawing

plateaukao:
In Undo example: https://jsfiddle.net/szimek/osenxvjc/
If pen's maxWidth is changed for each strokes, the final drawing of fromData() will all use the last stroke size.

**Solution:**
add current maxWidth to IPointGroup, and restore it for each saved stroke.

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -27,6 +27,7 @@ export interface IOptions {
 }
 
 export interface IPointGroup {
+  maxWidth: number;
   color: string;
   points: IBasicPoint[];
 }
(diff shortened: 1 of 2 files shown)


joshbmarshall:
Would be great if this would also save current minWidth

UziTech:
closing as stale

joshbmarshall:
That's unfortunate that this was stale. It means that I (still) can't store the drawing as a vector with different pen sizes, as loading the drawing up uses the last size pen. I've had to workaround by storing as an image and further edits write over top of a raster.

UziTech:
this was fixed in #571

PR #384, opened 2018-08-12
Retain images loaded with fromDataUrl.

pluby:
Plugs a small gap  with `fromDataUrl` and `toDataUrl`, allowing them to work like `fromData` and `toData`, although it doesn't populate the internal point structure so `toData` wont include the image, but I guess most people use either one format or the other.


UziTech:
closing as stale

PR #395, opened 2018-09-26
Update app.js

samirbhogayta:
Need to call save signature from asp.net .aspx page.


samirbhogayta:
Need to call save signature from .aspx page using asp button.

PR #404, opened 2018-11-02
Ability to maintain aspect ratio for fromDataURL

garhbod:
Haven't tested but was extracted from my working javascript

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -117,18 +117,32 @@ export default class SignaturePad {
 
   public fromDataURL(
     dataUrl: string,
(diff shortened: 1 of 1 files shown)


Zigonja:
Will this ever be implemented?

UziTech:
closing as stale

garhbod:
Shame this wasn't ever considered. Or is there another implementation of this?

UziTech:
This seems like something that could be handled outside of this library. They can calculate the ratio and pass that in. We want to keep this library small and not add too many more options.

PR #410, opened 2018-12-03
Use mjs instead of m.js

banyan:
refs: 

* https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#32-determining-if-source-is-an-es-module
* https://developers.google.com/web/fundamentals/primers/modules

mjs extention might be better for ES Module extention. what do you think?


UziTech:
closing as stale

PR #419, opened 2019-02-04
fixing split error

bellaj:
using "" instead '' in split will lead to an error :" cannot convert from string to char"


rickychilcott:
This should be merged.

PR #438, opened 2019-05-23
Allow disabling event handlers in the constructor

jmschneider:
Skipping the `on()` function allows me to use the [node canvas](https://github.com/Automattic/node-canvas) library to generate images from JSON on the server.

Something like this:

```js
const fs = require('fs')
const { createCanvas } = require('canvas')
const SignaturePad = require('signature_pad')
const canvas = createCanvas(200, 200)

let signaturePad = new SignaturePad(canvas, { off: true })
signaturePad.fromData(data)

const out = fs.createWriteStream('./test.png')
const stream = canvas.createPNGStream()
stream.pipe(out)
out.on('finish', () =>  console.log('The signature was written to a PNG file'))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ signaturePad.on();
 <dd>(string) Color used to draw the lines. Can be any color format accepted by <code>context.fillStyle</code>. Defaults to <code>"black"</code>.</dd>
 <dt>velocityFilterWeight</dt>
 <dd>(float) Weight used to modify new velocity based on the previous velocity. Defaults to <code>0.7</code>.</dd>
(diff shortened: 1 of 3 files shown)


UziTech:
closing as stale

jmschneider:
@UziTech I think this could still be a good improvement. I'd be happy to make any adjustments to the PR if needed.

UziTech:
I think we want to keep this a front-end library. There are a few other things we would need to make this actually work completely in node.

If you are just looking to create an image from data points maybe there is some way we could provide a static method to do that.

PR #445, opened 2019-07-05
Added toFile

manix:
Using the toFile method you can easily attach the signature to a FormData instance and avoid parsing.


UziTech:
We are trying to keep signature_pad limited in scope to drawing the signature and since this is already possible without needing access to private methods I don't think we will implement this change.

Issue #377, opened 2018-07-29
Signature disappear when zooming in with both fingers on Edge tablet


benag:
After signing when the user zoom in using touch on tablet running Edge browser, the signature disappear.

szimek:
I don't have a tablet with Edge, but I suspect you're updating size of the canvas on `resize` event. It's possible that zooming in is triggering this event and that's why the canvas is cleared. You might try to listen to [orientationchange](https://developer.mozilla.org/en-US/docs/Web/Events/orientationchange#Example) event instead, or disable zooming on the page with the canvas.

benag:
yes i am updating canvas size on window resize why would it clear the canvas?

Thanks

szimek:
It's how canvas element works - changing its "internal" size via `height` or `width` clears the whole thing. There are many issues about the same subject already.

The main solutions are - switch to `orientationchange` event or store content of the canvas changing canvas size in `resize` handler and then draw stored image after resizing, though it might be blurry.

benag:
Thanks, I will go for option 3

shalonteoh:
switched to orientation change event the pad still appear weird when sign although it doesnt clear signature anymore.

The below method works for me.

`
var cacheOrientation;

function resizeCanvas() {
  if (cacheOrientation != window.orientation) {
    cacheOrientation = window.orientation;
    var ratio = Math.max(window.devicePixelRatio || 1, 1);

    canvas.width = canvas.offsetWidth * ratio;
    canvas.height = canvas.offsetHeight * ratio;
    canvas.getContext("2d").scale(ratio, ratio);
    signaturePad.clear();
  }
}

window.onresize = resizeCanvas;
resizeCanvas();`

PR #446, opened 2019-07-11
allow canvas to be resized by CSS

jelhan:
This adds support for resizing the canvas used for signature pad by CSS.

Currently the library requires the drawingbuffer size of a canvas to equal it's rendered size. This is a hard limitation cause the drawingbuffer size determines the size of the image returned by `toBlob()` and `toDataURL()`. So basically it wasn't possible yet to generate an image with a different size as the signature pad was rendered with.

A simplified example of an use case that wasn't supported so far is:

```html
<canvas width="1000" height="500" style="width: 250px; height: 100px"></canvas>
```

This creates an image with a dimension of `1000x500px` but renders it on `250x100px`. This is especially useful if application requires fixed dimensions for the signature image created but the canvas should be responsive. It's supported out of the box after this change cause the scaling of the canvas is taken into account when mapping an element position to a position on canvas drawingbuffer.

While working on that feature I noticed that this also fixes the need for the resize hack. If I didn't missed something the resize hack was only necessary cause this library has taken the assumption that rendered canvas size and drawingbuffer size are equal. That assumption may have been wrong after resize (and in several other scenarios) causing the stroke to be drawn at a wrong position. So the recommended work-a-round was to change drawingbuffer size on resize events. This shouldn't be required anymore.

Some might to see it as a drawback that the rendered signature pad and the created image, which is returned by `toBlob()` and `toDataURL()` may have different ratios. That's the case as soon as drawingbuffer size and rendered size have different ratios. E.g. for `<canvas width="1000" height="500" style="width: 1000px; height: 250px;"></canvas>` the element would have a rendered ratio of `2:1` (`1000x500px`) but the generated image would have a `4:1` (`1000x250px`). Before the usage of signature pad was broken in that cases, so I wouldn't consider it a breaking change but more like a bug fix.


xiaoxiaofan123:
Hi I have the same problem.I use the library in the mobile,when i draw something in the canvas,and change the size of the canvas,the things i drew will be clear,how can i keep the things when i change the size.

jelhan:
@xiaoxiaofan123 This PR should fix your issue. Did you had a change to test it?

xiaoxiaofan123:
> @xiaoxiaofan123 This PR should fix your issue. Did you had a change to test it?

I use the version  "https://cdn.jsdelivr.net/npm/signature_pad@3.0.0-beta.3/dist/signature_pad.min.js"
But it seems still has problem,when i set the canvas.width='...', the signature will be cleaned.
I see the author said "It's how canvas element works"

https://github.com/szimek/signature_pad/issues/377#issuecomment-408673935

jelhan:
It should be fixed by this PR but this one is not merged and not included in any PR. I've pushed a build of this PR to my fork in `build` branch: https://github.com/jelhan/signature_pad/tree/build/dist You could use that ones to test it out. You should be able to use it directly in a script tag with `raw.githack.com` service: https://raw.githack.com/jelhan/signature_pad/build/dist/signature_pad.min.js

xiaoxiaofan123:
> It should be fixed by this PR but this one is not merged and not included in any PR. I've pushed a build of this PR to my fork in `build` branch: https://github.com/jelhan/signature_pad/tree/build/dist You could use that ones to test it out. You should be able to use it directly in a script tag with `raw.githack.com` service: https://raw.githack.com/jelhan/signature_pad/build/dist/signature_pad.min.js

I tried,the issue is fixed;
Have a new question,It was supposed to use "window.devicePixelRatio" to change the size of canvas,but it's no need since I use your code?

moshenskyDV:
@szimek any chance to merge this PR?

docs/js/signature_pad.umd.js
```diff
@@ -358,7 +358,19 @@
       };
       SignaturePad.prototype._createPoint = function (x, y) {
           var rect = this.canvas.getBoundingClientRect();
-          return new Point(x - rect.left, y - rect.top, new Date().getTime());
+          var positionOnCanvasElement = {
+              x: (x - rect.left),
+              y: (y - rect.top)
+          };
+          var ratio = {
```
bpetty-formfast, line 365: This object will never change, correct?  It would be more efficient if it wasn't calculated on every call to_createPoint(...).

bpetty-formfast:
@jelhan, I noticed that when our code scales our canvas using a CSS Transform/Scale(...), I have to apply a work around similar to the "hack" mentioned in the readme (but using my own scale value, not devicePixelRatio).  That said, it kind of works but the stroke doesn't scale... it stays a consistent thickness.  Does your patch scale the stroke thickness with the zoom level?

docs/js/signature_pad.umd.js
jelhan: It's quiet a long time since I implemented this one. So I'm not :100:ly sure to be honest. I fear that in some edge cases window resize or zoon could alter `clientWidth` or `clientHeight` of the `canvas`.

I don't think the performance difference is significant. We execute `getBoundingClientRect()` on `canvas` anyways. So the browser would calculate styles and do the layout even if the object would be cached.

I tend to keep it for now and refactor to a cached solution later. Don't want to have too much risky changes in one pull request. Especially as there are no tests if I recall correctly.

jelhan:
>  Does your patch scale the stroke thickness with the zoom level?

Good question. Not sure to be honest. Will try to test within the next days. If you have time earlier to try it out, that would be great!

emilfihlman:
If I'm not mistaken, this breaks a lot of stuff and is just wrong.

Canvas width and height are completely different things from the style width and height. Correct behaviour is to:
Respect the actual pixel dimensions (width and height). If the developer wants to draw the canvas stretched then that's on the developer, but don't ever go adjusting the width and height.

get mouseover location of the element with clientX and clientY and adjust the offset by getBoundingClientRect top and left.

Then you scale by the relative positions to draw.

UziTech:
closing as stale

jelhan:
@UziTech Has this been fixed in another pull request? The merge request is stale due to missing review. I'm willing to work on it if there is a chance to get it in.

UziTech:
@jelhan see https://github.com/szimek/signature_pad/pull/446#issuecomment-777126999

I don't believe taking away the ability to have different canvas dimensions and style dimensions is necessary.

jelhan:
> I don't believe taking away the ability to have different canvas dimensions and style dimensions is necessary.

@UziTech I'm a little bit confused. The goal of this pull request is to add support for canvas dimensions that do not match style dimensions. This is still not supported using the latest release as shown in this JSFiddle: https://jsfiddle.net/m8ywqf6d/1/

Are you willing to support that use case? It is required by the project, which I'm using this library in. My project requires that the generated images has a fixed width and height in pixel. At the same time the canvas element used to draw that image should fill up all available width. Simplified it looks like this: `<canvas width="500" height="250" style="width: 100%"></canvas>`.

If you are not open to land such a feature, I will continue to use my feature branch. It is working fine so far.

UziTech:
Ah yes, I was confused by the comment above. I thought you were trying to make it so the style width was always the same as the canvas width.

UziTech:
Could you rebase and maybe write a test for thise?

jelhan:
Rebased on latest master. Will try to write a test later. I can not recall why I haven't written one when I created this merge request.

UziTech:
@jelhan are you still working on this?

jelhan:
> @jelhan are you still working on this?

I didn't had time to actively work on it. But it is still on my todo list.

As far as I recall two things are left:

1. Ensure that stroke width also works well in a resized canvas.
2. Add tests for it based on the example setup in #576.

UziTech:
@jelhan I think this is good to go. If you want to look it over and approve I can merge it.

UziTech:
It seems the biggest issue with doing this is that the image saved from `toDataURL()` will be the same size as the canvas width and height not the size it looks on the page.

size on page (css size):
image

canvas and image size:
image

jelhan:
> It seems the biggest issue with doing this is that the image saved from `toDataURL()` will be the same size as the canvas width and height not the size it looks on the page.

That's s feature in my perspective. At least for my use case I needed to collect a signature as an image with specific size (e.g. 600x300px). But the input should be responsive in the sense that it scales with device screen size.

emilfihlman:
> > It seems the biggest issue with doing this is that the image saved from `toDataURL()` will be the same size as the canvas width and height not the size it looks on the page.
> 
> That's s feature in my perspective. At least for my use case I needed to collect a signature as an image with specific size (e.g. 600x300px). But the input should be responsive in the sense that it scales with device screen size.

With respect, that's no longer a signature since you are editing it. Signature must always match what the user saw when writing it.

And it's also wrong. If you want to fit a signature onto some certain size then you use a contain algorithm, which is super easy.

emilfihlman:
> It seems the biggest issue with doing this is that the image saved from `toDataURL()` will be the same size as the canvas width and height not the size it looks on the page.
> 
> size on page (css size): image
> 
> canvas and image size: image

You can't allow a merge with this issue.

jelhan:
> With respect, that's no longer a signature since you are editing it. Signature must always match what the user saw when writing it.

Just to be clear: We are _only_ talking about rescaling it without affecting dimensions, stroke size or anything else. At least that would be the goal for this issue.

You can not guarantee that the scaling of the signature is the same when user enters it as well as when it is viewed or printed. That depends on many factors. Most importantly specifics of the hardware used like DPI.

jelhan:
Not sure if you are talking about CSS which affects dimensions. E.g. by having a canvas of `100x100px` but stretching it with CSS to `200x100px`. I would consider that scenario as invalid or buggy usage of the library. Not an issue of the library itself.

emilfihlman:
Signatures must retain their aspect ratios at all times and imagedata should return a signature that respects that.

This patch is just wrong and this has already been indicated in the beginning. The correct way to go about having a signature area that fits into a given area responsively is to either a) initalise the size when we know the area we have (the best choice for static content), or b) rescale the canvas (and redraw the existing content) to fit the area when we get a resize event (the best choice if we need full responsiveness), or c) to use cover as an algorithm and only showing a subsection of the drawable canvas (less code, but imho a stupid decision).

After that if for some reason you need to fit a signature into a known pixel box, you take the signature and contain it there, possibly preferably after trimming any whitespace around it which you should anyways do. The library could easily provide functions for these.

The most stupid idea is to allow changes in aspect ratios between representation and actual underlying data. Nothing good comes out of that.

Finally, all these are easy to implement, I've done the hardest (second one) for another project. There is a single design decision there to be made regarding existing content, one can either crop, contain or have a scrollable drawing area (which should probably resize if cleared).

UziTech:
> Signatures must retain their aspect ratios at all times and imagedata should return a signature that respects that.

Can you share references for this statement? As far as I know (in the USA at least) the intent to sign is all the user needs. What the actual signature looks like is not important.

I kind of agree with @jelhan this PR does not prevent people still using it the old way and keeping the canvas size the same as the as size. It just adds the ability for them to be different if the user of this library wants to and is ok with the canvas size and css size being different.

emilfihlman:
If the intent to sign is all it takes then a button is enough, and all transformations on the signature are valid.

This is absolutely not the case.

If you do go about accepting this lunacy that could easily be handled in a much better way, you must provide apis to normalise the signatures so that application writers can get proper signatures out of the library. Though I suspect you'll find it much easier to just do what I said. Not to mention that one can already use object-fit: cover to immediately get the wanted result without messing with aspect ratios.

As a final warning, if you implement this patch as is it prevents people from reusing the canvas easily after stopping the signature since the aspect ratios are broken.

UziTech:
@emilfihlman if you would like to create a PR showing a better way of handling this I would be happy to accept it.

This doesn't prevent anyone using the canvas the same way they are currently (keeping the same canvas and css size). If your concern is that users may not know they need to keep the same size for the image to be the size that is displayed we can add a note to the readme.

UziTech:
@szimek what do you think about this? (read from https://github.com/szimek/signature_pad/pull/446#issuecomment-1059888995) If you are ok with it I can merge this in. I did add a note about keeping the canvas and css size ratios the same or the image will be distorted.

UziTech:
maybe we could even add this as an option so it won't be a breaking change

Waleed-KH:
I was able to make the canvas responsive and CSS resizable while retaining the signature aspect ratios by using `ResizeObserver`, I also used an event listener for the `resolution` media query to update the canvas scale (instead of window `resize`)

Recording 2022-07-30 at 03 33 38

https://jsfiddle.net/0u5ofkc3/

UziTech:
I'm going to close this because it does seem that there are easy ways to accomplish this without the breaking changes.

Waleed-KH:
yes, but a lot of devs don't know about it, and results to unexpected behaviors, many of the reported issues are related to this problem [#638, #618, #639, #600 ...] so I think there should be a **built-in** way to solve this

UziTech:
We could make it more prominent in the readme. If devs don't read the readme than it won't be easy to explain the breaking changes if this was merged either.

PR #448, opened 2019-07-16
creating time based signature draw

eliceguipedro:


eliceguipedro:
This method allow to recreate the signature just like fromData();
but this is time based draw, takes in consideration the time between points and strokes.

I copy the function fromData and _fromData to create
fromDataDraw and _fromDataDraw, maybe its better to merge the funcionality in one unique method,
something like
signaturePad.fromData(data,timebased)

UziTech:
closing as stale

PR #461, opened 2019-11-03
Added functionality similar to addFromData from v2

johnventions:
This change adds the "addFromData" function, which adds points to the SignaturePad without clearing the canvas first. This feature existed in v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ const data = signaturePad.toData();
 // Draws signature image from an array of point groups
 signaturePad.fromData(data);
 
+// Draws signature image from an array of point groups, without clearing your existing image
+signaturePad.addFromData(data);
+
 // Clears the canvas
 signaturePad.clear();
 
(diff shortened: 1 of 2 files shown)


johnventions:
Initial PR failed codelimate due to duplicate code. I have updated the code to instead have an optional `clear` parameter on the "fromData" function which defaults to true. If set to false, the signatuePad will not clear the drawing before adding the points.

UziTech:
Can you rebase and add a test?

@szimek I think this could be useful for using `fromData` on a canvas that has a background image.

szimek:
@UziTech Yeah, there were a few tickets related to using background images. One thing - I'd probably change boolean to options (`false` => `{ clear: false }`) just in case we ever add more options there, though I don't really have a strong opinion about it.

UziTech:
I like the options object better as well. :+1:

@johnventions is this something you are still willing to work on? or I could take it over from here.

johnventions:
@UziTech - all yours if you want to jump on it :)

I'm a bit disconnected from the project at the moment

PR #463, opened 2019-11-07
Undo feature as part of base functionality

johnventions:
The examples show a method of how to do an undo process, but you're required to manually modify the data array and redraw it. This update puts the undo function into the signature pad as a base method and modifies the data array to track brush sizes throughout the drawing process so that brush min/max is not lost during the re-draw process.


mementototem:
You need to

- Edit `/tests/fixtures/face.ts` file to match new `PointGroup` interface
- Change interface name to match beta-4 version

UziTech:
We are trying to keep signature_pad small and prevent feature creep. This seems easy enough to do without changing signature_pad.

PR #477, opened 2020-04-23
Feature/signature from text

arunredhu:
- Feature added to signature from a text string
- Developer can now provide a text either may be from a text input or any other source to signature along with font style.


UziTech:
I don't think this should be part of signature_pad. We want to keep this library focused on drawing the signature.

PR #481, opened 2020-04-28
Fix #480 - Fix race condition / edge case in _strokeUpdate where this._data can sometimes be []

ndbroadbent:
Fixes #480 

Fixes this error that I'm getting sometimes:

Screen Shot 2020-04-28 at 10 09 48 PM

This can happen if `clear()` is called while a signature is still being drawn, so the signature might need to be "restarted" by calling `_strokeBegin` again.

PR #495, opened 2020-08-21
fix(signature_pad): fix race condition in _strokeUpdate

koprivajakub:
Since there is no initiative to release the patch on version 2.x.x I am trying to escalate this by creating the same patch from #481 into branch for major version `2`.

@szimek can you please take a look on this and release a patch version? Thank you very much.


koprivajakub:
@szimek Any idea whether this will get merged and released as patch to version `2.x.x` ?

koprivajakub:
@szimek any intention to merge this?

UziTech:
this will be fixed in v4 released soon.

PR #501, opened 2020-09-27
Fix(msTouch)

ronaldoasevedo:
Fix Error Below: 
14:9 Subsequent property declarations must have the same type.  Property 'msTouchAction' must be of type 'string | null', but here has type 'string'.
    12 | declare global {
    13 |     interface CSSStyleDeclaration {
  > 14 |         msTouchAction: string;
       |         ^
    15 |     }
    16 | }
    17 | export interface Options {

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -15,7 +15,7 @@ import { throttle } from './throttle';
 
 declare global {
   interface CSSStyleDeclaration {
-    msTouchAction: string;
+    msTouchAction: string | null;
   }
 }
 


szimek:
Hi @ronaldoasevedo! When do you get this error? When I build the project or run linter I don't get any errors.

jcuadrosj:
Hi @szimek! I'm seeing the same error after updating from `3.0.0-beta.3` to `3.0.0-beta.4`. Getting the error on compile time, with no other changes. Previously worked without issues.

szimek:
@jcuadrosj Which version of TypeScript are you using? Is it the one in the lock file, 4.0.3? Or maybe 4.1.x beta? I'm not getting this error, so it's really strange...

The error ("Subsequent property declarations must have the same type...") suggests that it's already declared somewhere else, but I added this declaration exactly because TS 4.0.3 was complaining that `msTouchAction` is not declared anywhere.

jcuadrosj:
@szimek  That might be it. We're using TS 3.7.3.

szimek:
I guess it would then make sense to merge this PR. If someone uses TS 4.x, it would be declared for the first time, if they use TS 3.x it would be redeclared.

@jcuadrosj Does this PR fix this issue for you? If so, I'll merge it.

jcuadrosj:
Yes, this change does fix the issue for me. Thanks @szimek!

szimek:
Thanks @ronaldoasevedo!

PR #503, opened 2020-10-09
Capture pressure when available and implemented a function to convert the captured data to the ISO 19794-7 specification. #213

aegeavaz:


kurtuluso:
Will it be merged?

PR #506, opened 2020-10-21
Create composer.json

matteobaccan:
packagist.org integration


socieboy:
why he would do that? this is a Javascript repository.

UziTech:
closing as stale. I don't think we need to maintain another package manager.

Issue #520, opened 2020-12-14
Events doesn't work after Init on IOS Chrome / Safari


alekradev:
**Do you want to request a feature or report a bug?**
report a bug

**What is the current behavior?**
After the initialization, Click or Touch events no longer work under IOS

**What is the expected behavior?**
That the events still work after initialization

**Which versions of SignaturePad, and which browser / device are affected by this issue? Did this work in previous versions of SignaturePad?**
Chrome / Safari on IOS

commi:
The signature-pad also interferes with angular-click-handlers. They can only be restored by calling .off(). This PR fixes this.

PR #521, opened 2020-12-14
Events should only be handled within the canvas. Under IOS this leads…

alekradev:
Events should only be handled within the canvas. Under IOS this leads to problems accessing elements and changing event listener there.

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -167,11 +167,11 @@ export default class SignaturePad {
 
     this.canvas.removeEventListener('pointerdown', this._handleMouseDown);
     this.canvas.removeEventListener('pointermove', this._handleMouseMove);
-    document.removeEventListener('pointerup', this._handleMouseUp);
+    this.canvas.removeEventListener('pointerup', this._handleMouseUp);
 
     this.canvas.removeEventListener('mousedown', this._handleMouseDown);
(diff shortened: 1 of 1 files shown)


alekradev:
Ich hatte ein Issue erstellt siehe #520

cklm (approved):
fixes the problem for me

UziTech:
The problem with having the up events only on the canvas is when the user presses the mouse button on the canvas, moves the cursor off the canvas, then releases the mouse button signature pad doesn't know that the button was released so when the user moves the cursor back on the canvas signature pad doesn't know the button was released and will continue drawing the line without the button pressed.

I don't think this PR will be merged because of that.

cklm:
Sorry, can't agree here - this is a serious problem due to the glitches of iOS/iPadOS. You described it right, but on apple-devices that happens often. We have encountered the problem in daily use, this PR fixes it (at least in our case). So I would be glad, if everyone could benefit from it.

szimek:
@UziTech explained the reason why currently all "up" events are handled on the document rather than canvas.

@alekradev @cklm could one of you explain what the problem is exactly, preferably with an example? If we understand better what the issue is, maybe we can handle it differently only on iOS, but we can't merge this PR as it is.

PR #528, opened 2021-03-12
use getCoalescedEvents when it is available

jwchang0206:
I believe that using `getCoalescedEvents` will get much precise points for drawings, if the `pointermove` is available on a user's browser.

Reference: https://developers.google.com/web/updates/2017/06/aligning-input-events

---

[Current "Signature_pad" demo](http://szimek.github.io/signature_pad/)

vs

[getCoalescedEvents demo](https://codesandbox.io/s/pointermove-drawing-omwnk?from-embed)


jwchang0206:
@szimek Would this be mergeable?

UziTech:
@jwchang0206 could you rebase this?

UziTech:
`getCoalescedEvents` is experimental so I don't think we want to use it yet.

PR #530, opened 2021-03-26
Added toSkiaPath and fromSkiaPath public methods

philippjbauer:
For your consideration:

I needed the option to create a SkiaPath representation from the signature_pad for usage in SkiaSharp.

The format is not dissimilar from SVG paths. I.e., HEX_COLOR;SVG_PATH_STRING;HEX_COLOR;SVG_PATH_STRING;...

I used the existing path construction method found in the _toSvg method to create the string for toSkiaPath. And the existing import capabilities for SVG to realize the fromSkiaPath functionality.


UziTech:
I don't feel like this should be part of the library. Are there any private methods needed to do this outside of this library?

PR #538, opened 2021-05-16
Allow offsets when loading image via fromDataURL

siskavera:
Allow x and y offset to enable centering an image with different dimensions from the canvas.

This is not quite as good as an automatic version which would scale & fit the image to the canvas, but at least it allows all the necessary manual settings.


siskavera:
@szimek or anyone else, could you take a look?

PR #541, opened 2021-06-03
feat: add toSVG

UziTech:
- update deps
- update scripts to work on windows
- add `.toSVG` function to create an SVG element from the data and trim it so it is just as big as the signature.


UziTech:
I can break this into multiple PRs if you want but I thought I would put it out there first.

@szimek I noticed the Note on the readme for maintainers. I would be happy to help maintain this as I use it in a few projects.

Sparticuz:
Any particular reason why trim was never merged?

UziTech:
Because there is already better ways to do it https://github.com/szimek/signature_pad#removing-empty-space-around-a-signature

Issue #399, opened 2018-10-12
support cross-origin fromDataURL


somazx:
_If you have a question, please ask it on [StackOverflow](https://stackoverflow.com/questions/ask?tags=javascript+signature-pad) and use `signature-pad` tag instead of creating an issue._

**Do you want to request a feature or report a bug?**
Add support for loading signature Cross Origin with `fromDataURL`.

**What is the current behavior?**

In Firefox (possibly/likely other browsers as well) if I attempt to load a signature image from a domain other than the origin (for example if the signatures are stored on S3/cloudflare) then future attempts to call toDataURL result in a SecurityError due to "tainted canvas", as explained here https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image

_If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via <https://jsfiddle.net> or similar (template: <https://jsfiddle.net/szimek/h80phxyr/>)._

https://jsfiddle.net/ab27n8f3/1/

Then click the save as PNG or JPG buttons and note the error in the console.

`SecurityError: The operation is insecure.` (Firefox)

Chrome:
```
signature_pad.min.js:1 Uncaught DOMException: Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported.
    at o.toDataURL (https://cdn.jsdelivr.net/npm/signature_pad@2.3.2/dist/signature_pad.min.js:1:3363)
    at HTMLButtonElement.<anonymous> (https://fiddle.jshell.net/ab27n8f3/1/show/:120:27)
```

**What is the expected behavior?**

Support cross origin requests for image signatures

**Which versions of SignaturePad, and which browser / device are affected by this issue? Did this work in previous versions of SignaturePad?**

Signature Pad 2.3.2
Firefox 62.0.3

I believe the solution is to add just one line to `fromDataURL`, after:

`const image = new Image();`

add the following after:

`image.crossOrigin = "Anonymous";`

jogi323:
+1
I am also facing the same issue after clear existing  signature and trying to add new signature. @somazx   solution is working locally but need it from signature_pad package.

szimek:
Fixed by #542.

PR #542, opened 2021-06-04
Added Anonymous to crossOrigin prop

pxwee5:
Discussion: https://github.com/szimek/signature_pad/issues/399
Tested and it works for my setup. What do we need to get this through ? @szimek

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -138,6 +138,7 @@ export default class SignaturePad {
         callback(error);
       }
     };
+    image.crossOrigin = 'Anonymous';
     image.src = dataUrl;
 
     this._isEmpty = false;
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -138,7 +138,7 @@ export default class SignaturePad {
         callback(error);
       }
     };
-    image.crossOrigin = 'Anonymous';
+    image.crossOrigin = 'anonymous';
     image.src = dataUrl;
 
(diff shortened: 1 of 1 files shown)


src/signature_pad.ts
```diff
@@ -138,6 +138,7 @@ export default class SignaturePad {
         callback(error);
       }
     };
+    image.crossOrigin = 'Anonymous';
```
UziTech, line 141: Not sure if it matters but the spec shows it as lowercase. [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)
```suggestion
    image.crossOrigin = 'anonymous';
```

UziTech (changes requested):

UziTech (approved):

szimek (approved):

pxwee5:
Any plans to do a patch release in the 2.x.x branch ?

szimek:
@pxwee5 Yeah, thought it's really hard to say when exactly. Do you have a list of bug fixes that were merged into 3.x branch that you'd like to have in 2.x release? This would help a lot!

pxwee5:
Hey @szimek. I'm not aware of any other bug that was fixed in v3 that needs to be in v2. If there are any that you think should fix v2 we should definitely list them out. 
But in my opinion, the `crossOrigin` fix alone is critical enough to warrant a patch fix for v2.

PR #546, opened 2021-06-19
fix _drawCurve() when moving very slowly

marmooo:
On Android devices, curve.length() may return a number less than 1.
In other words, if you write a line very very slowly, 
Math.floor() will set the length of the line to 0.
This leads to a bug where the line is not written at all.
This bug can be solved by not using Math.floor().

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -419,7 +419,7 @@ export default class SignaturePad {
     const widthDelta = curve.endWidth - curve.startWidth;
     // '2' is just an arbitrary number here. If only lenght is used, then
     // there are gaps between curve segments :/
-    const drawSteps = Math.floor(curve.length()) * 2;
+    const drawSteps = curve.length() * 2;
 
     ctx.beginPath();
     ctx.fillStyle = color;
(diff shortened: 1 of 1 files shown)


UziTech:
I can't seem to reproduce this issue on the [demo](http://szimek.github.io/signature_pad/) site. I tried it on a pixel 5 with android 11.

Can you provide a screenshot or video of the issue?

marmooo:
I have encountered this problem when using MediaPad M3 Lite 10 wp (Android 7.0).
I guess this problem to only happen on certain Android devices, 
because each product has a different implementation of the touch panel.

https://user-images.githubusercontent.com/28778641/123040746-34d72b00-d42f-11eb-9791-767392bbbaa9.mp4

szimek:
While not stated explicitly, `drawSteps` is supposed to be an integer. That's why I think that replacing `Math.floor` with `Math.ceil`, rather than removing it completely, might be a better solution.

marmooo:
After receiving your thankful advice, I checked [the definition of canvas in HTML Living Standard](https://html.spec.whatwg.org/multipage/canvas.html#2dcontext).
As a result, I report that `ctx.arc` used in `drawSteps` was defined as `double` type.
If you follow the definition, the solution without `Math.floor` looks correct. 
Both the solution without `Math.floor` and the solution with `Math.ceil` work, so I'll follow the judgment of the review. 
I haven't found any issues with either implementation at this time.
Which fix is better?  I will modify the code according to the future replies.

szimek:
I'd go with `Math.ceil`. `drawSteps` name suggests number of steps (or iterations), so it makes sense for it to be an integer. E.g. with 1.8 iterations it's not obvious if it's actually 1 or 2 iterations.

Issue #559, opened 2021-10-13
fromData() and Point class problem with string coords


hembachrterran:
**Do you want to request a feature or report a bug?**
Report a bug

**What is the current behavior?**
If fromData() function get a PointsGroup and the x and y coords are string (e.g. "54.242422523") the type is not checked that causing 'NaN' values in Bezier class calculateControlPoints() function and draw a wrong empty drawing.

**What is the expected behavior?**
Check for type, if string then parse float

**Which versions of SignaturePad, and which browser / device are affected by this issue? Did this work in previous versions of SignaturePad?**
v3.0.0-beta.4

**Suggestion to fix**
`class Point {
    constructor(x, y, time) {
        this.x = ((typeof x == 'string') || (x instanceof String)) ? parseFloat(x) : x;
        this.y = ((typeof y == 'string') || (y instanceof String)) ? parseFloat(y) : y;
        this.time = time || Date.now();
    }`

romarioliveira25:
> @UziTech Yes, I will in the next few days.

Any news about that?

hembachrterran:
@romarioliveira25 Sorry, I have no time at the past days. Today I will make a PR to fix this problem.

PR #563, opened 2021-11-07
Fix string type

Hembi:
Related to:
https://github.com/szimek/signature_pad/issues/559

### DIFF
--- a/docs/js/signature_pad.umd.js
+++ b/docs/js/signature_pad.umd.js
@@ -11,8 +11,8 @@
 
     var Point = (function () {
         function Point(x, y, time) {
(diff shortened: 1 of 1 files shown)


UziTech:
You should change [`/src/signature_pad.ts`](https://github.com/szimek/signature_pad/blob/master/src/signature_pad.ts) not `/docs/js/signature_pad.umd.js`

UziTech (changes requested):
You should change [`/src/signature_pad.ts`](https://github.com/szimek/signature_pad/blob/master/src/signature_pad.ts) not `/docs/js/signature_pad.umd.js`

PR #566, opened 2021-11-14
Capture pressure when signing

### LINKED ISSUE #213: Capturing biometric data with the sign
Hello.

Is it possible to get biometric data associated to the sign? I say, writing speed, number of lines, etc. I need to be sure that the user doesn't reject the authenticity of his own sign for legal purposes. This component is the only one I've seen that writes a smooth sign, so I suppose that data is getting captured. The end goal is getting the end user sign + biometric information associated to that sign + legal third-party timestamp; then, encoding the whole pack, and send to the database. Is it possible anyway?

Thanks a lot.

### LINKED ISSUE #379: Feature request: Respect pressure/force on touchevents
When using for example an Apple Pen, I would love if signature_pad could respect the pressure/force property of a touch event to adjust the dotSize dynamically. Is this a feature that could be easily implemented somehow?

I will experiment a bit myself and see if I come up with a solution. Do you have any pointers of where this functionality could be implemented @szimek?

One idea is to pass the mouse/touch event to the dotSize option whenever this is a function, and let that function decide what to do with it. Would this work?

UziTech:
update of #503

closes #503
closes #213
closes #379


szimek (approved):
I hope it works ;) 

I'm wondering if we could make some of the properties (e.g. `_data`) public, or rather stop pretending they are private, or provide getters for them, and document them properly. This would allow us to create `toWhateverFormat`-type function either as a separate function (e.g. `toSvg(signaturePad.data)`, `toIsoData(signaturePad.data)` that wouldn't have to be loaded if not used, or even as a third-party add-on library.

Also, I just realized that if we stopped supporting Safari 12 and earlier, we could handle pointer events only... However, it might still be too early - I don't know what types of projects this library is used in.

UziTech:
`_data` is kind of public with the `toData()` method. I was thinking about that too. Instead of having all of these `to*` methods just making them separate packages.

Should I remove the `toISOData` method and let that be a separate package?

szimek:
I forgot about `toData()` 🤦 .

I think it would be a good idea to extract it. I'm just wondering about API - if these `toWhatever` functions should accept the whole signaturePad instance or just the result of `toData()`. E.g. `toSVG` (which theoretically could also be extracted to a separate function, unless we treat it as a core feature) uses some additional properties like `dotSize`.

UziTech:
One of my next changes was going to put more of those properties into the point groups so we can save information like max/min width and dot size, etc.

I feel like toData() should give all of the properties.

PR #567, opened 2021-11-15
use event emitter

### LINKED ISSUE #543: Make signature_pad an event emitter
This is an issue to track https://github.com/szimek/signature_pad/pull/353#issuecomment-857077908 probably for v4

UziTech:
- Remove callback for fromDataUrl and return promise instead.
- Remove `onBegin` option and add `beginStroke` event.
- Remove `onEnd` options and add `endStroke` event.
- add `beforeUpdateStroke` and `afterUpdateStroke` events.

fixes #543 
closes #423

### DIFF
--- a/docs/js/signature_pad.umd.js
+++ b/docs/js/signature_pad.umd.js
@@ -316,11 +316,11 @@
             return this._data;
         };
         SignaturePad.prototype._strokeBegin = function (event) {
+            this.dispatchEvent(new CustomEvent('beginStroke', { detail: event }));
             var newPointGroup = {
                 color: this.penColor,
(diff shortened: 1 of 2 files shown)


szimek (approved):
Looks awesome! 🎉 

The only small issue is that this will break IE compatibility, but I think it's ok to do in a major release. I'm wondering if we should just skip 3.0 release given that the last 3.x beta was released in 2018 and go straight to 4.x.

UziTech:
Ya I don't think we should continue to support IE given Microsoft hasn't supported it in a while.

I think going to v4 would be fine.

PR #569, opened 2021-11-16
Reconstruct internal data from image when instantiating from image url

arcman7:


UziTech:
Can you write tests for this? I don't think it will work the way you think it will work.

arcman7:
Sure thing. But can you tell me what you would like to see that would give you confidence this works?

UziTech:
This seems to load points from left to right top to bottom. I don't think that will work because signature pad will draw lines between those points.

To get this to actually work you will need to somehow get the line segments and figure out which points will create the correct bezier curve. (I don't think this is actually possible).

If you can show that this actually works with a test showing that you can recreate a signature from the data points given by an image I would be very willing to admit I was wrong.

arcman7:
> This seems to load points from left to right top to bottom. I don't think that will work because signature pad will draw lines between those points.
> 
> To get this to actually work you will need to somehow get the line segments and figure out which points will create the correct bezier curve. (I don't think this is actually possible).
> 
> If you can show that this actually works with a test showing that you can recreate a signature from the data points given by an image I would be very willing to admit I was wrong.

Yeah I realized that the exact reconstruction would be impossible since your curves use averages. However, do we really need the original "source" points? I think for the majority of use-cases, it's simply good enough to construct PointGroups for the generated curves. No?

arcman7:
> What use-cases would that be useful? If you can't recreate the signature on the canvas from the points why do we need the points?

My expectation here is that from a saved signature (image data URL), I can reconstruct the x-y points that were generated by signpad. 

Why is this useful? Currently, we use this library to handle signing forms and whatnot, but also to validate signatures. This is where this feature becomes important. We need to confirm that signatures are not comprised of a single "dot" or a single line stroke. These are not valid signatures and our partners that we work with will not accept them. I started this feature by creating an algo to analyze signpads internal data structures, and to basically measure a "degree of complexity". This works well enough for new signatures, but for existing ones, we need to do the reconstruction. I figured there would be others that have this usecase as well and decided to make it a PR. It's your call really, I just thought it would be nice. Also, adding this method would really makes signpad work well with redux form validations.

UziTech:
> I can reconstruct the x-y points that were generated by signpad.

I can see how that would be useful but, as you stated before, that is not possible. Which is why I am questioning how this is useful.

arcman7:
> I can see how that would be useful but, as you stated before, that is not possible. Which is why I am questioning how this is useful.

Okay, I think I'm misunderstanding you here. Images generated by signpad have fictitious pixels inserted when using these fancy curves to produce a signature. When I refer to "source" points, I'm actually referring to the set of xy points captured from the user-generated stroke events used to generate the fancy curves. I'm saying I can't reconstruct the exact stroke movements the user took, but I can reproduce the fictitious pixels with the method I provided, no? This feature PR is only attempting to add behavior that allow signpad to initialize with an internal dataset when calling from `fromDataURL`. 

Are you saying that this won't work as it is?

UziTech:
> but I can reproduce the fictitious pixels with the method I provided, no? 

No, it looks like the method you provided just loads every pixel that is not transparent in the image from left to right, top to bottom.

If you could reproduce the "fictitious pixels" that could be useful. But I don't think you can without some image recognition model.

arcman7:
> > but I can reproduce the fictitious pixels with the method I provided, no?
> 
> No, it looks like the method you provided just loads every pixel that is not transparent in the image from left to right, top to bottom.
> 
> If you could reproduce the "fictitious pixels" that could be useful. But I don't think you can without some image recognition model.

Oh. Yeah, I did not intend this to ever be used with random images. Only images generated by signpad. You think this would happen a lot?

UziTech:
> Oh. Yeah, I did not intend this to ever be used with random images. Only images generated by signpad. You think this would happen a lot?

Yes. There would be a lot of people complaining that this does not work with their grainy images.

But I don't think the current method would even work with images produced by signature_pad because it is just listing pixels from left to right not in a line if the line is vertical like the below image

image

UziTech:
> IF I can demonstrate this feature and write tests, you would accept this PR?

Sure, I think a good test would be if you can call `fromDataURL` then `signaturePad.fromData(signaturePad.toData())` and get an image that looks similar to the original image.

If calling `signaturePad.fromData(signaturePad.toData())` doesn't produce a similar image I don't think the data is useful.

arcman7:
Sounds good. It'll be a little bit before I make this pr. I'll have to look at how your functions use velocity to figure out the expected direction a stroke may take. It might just boil down to a least-regression-model fit, but we'll see :)

UziTech:
I'm going to close this as there hasn't been any work on it in a while. Feel free to reopen it or start a new PR if you want to continue working on it.

PR #574, opened 2021-11-23
Release V4.0.0

UziTech:


szimek (approved):
Thank you so much!



I thought about switching to [microbundle ](https://github.com/developit/microbundle), which hopefully should provide better defaults than what we have now, but I really don't want to delay this release anymore. We can always release 5.0.0 sooner than 3 years from now :)

I can publish the new release to npm later today.

PR #575, opened 2021-11-23
Use touch events for iOS platforms instead of point events

### LINKED ISSUE #518: Signing with Apple Pencil is not possible
Bug

While signing with apple pencil not all lines are in the canvas

_If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via <https://jsfiddle.net> or similar (template: <https://jsfiddle.net/szimek/h80phxyr/>)._
With the iPad Air 2020 and Apple Pencil 1 just use the demo.
http://szimek.github.io/signature_pad/

All lines are shown in canvas.

Current Safari on iPad

sumbad:
The "Scribble" feature of iOS intercepts point events. So that we can lose some of them when tapping rapidly. We should use touch events for iOS platforms according to https://developer.apple.com/forums/thread/664108 and https://stackoverflow.com/questions/64314226/ipados-14-apple-pencil-fast-tapping-not-working-html-javascript-ontouchstart/64316781#64316781.
I've tested this solution on iPad Air 2020 with Apple Pencil. It works right.

closes #518

### DIFF
--- a/docs/js/signature_pad.umd.js
+++ b/docs/js/signature_pad.umd.js
@@ -139,7 +139,6 @@
         constructor(canvas, options = {}) {
             super();
             this.canvas = canvas;
-            this.options = options;
             this._handleMouseDown = (event) => {
                 if (event.buttons === 1) {
                     this._drawningStroke = true;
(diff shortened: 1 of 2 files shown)


UziTech:
Is there some other way to detect the pencil instead of changing this for anyone on a mac?

sumbad:
I was thinking about it but didn't find another solution.

It's only possible to detect the type of input from a user after an event was fired. For the Point Event there is `event.pointerType` and `event.touches[0].touchType` for the Touch Event. But we want to know about it before the first event. If I'm right it's not possible.

The code `/Macintosh/.test(navigator.userAgent) && 'ontouchstart' in document` returns `true` only for iOS, for MacOS this code returns `false` because there is no `ontouchstart` in the `document` object.

PR #576, opened 2021-11-23
add basic test which simulates user input

jelhan:
When trying to write a test for #446, I noticed that there aren't any test yet covering user interactions. After some iterations I came up with a possible solution, which at least works for simple cases. I thought it may make sense to open a separate pull request for that one to keep the discuss in #446 focused.

My original goal was to paint a simile in that test. But I faced some issues when trying to paint a stroke. So for the first iteration it is three dots only.


UziTech (approved):
Thanks!

jelhan:
Glad to hear that you are happy with that approach. Will use it to add a test for #446.

PR #581, opened 2021-12-08
Create adapter class for EventTarget so that it would work in older iOS.

bowozer:
It would check whether it could create EventTarget object. If failed,
then it use document object as its EventTarget object.

This is to support iOS 13 and older version.

If this needs anything else, please don't hesitate to let me know.


UziTech:
Can you write a test for this to make sure `signaturePad.on('beginStroke')` and other events work?

bowozer:
Added two commits.

11286e45378a78e09bf6a8762145ad8000b2fc8c contains unit tests that should have been added in PR https://github.com/szimek/signature_pad/pull/567, but for some reason those aren't.

40965b78f154ed5a67da04925299f40bcf4ef351 added a new option 'documentAsEventTarget' and unit tests to see whether the SignaturePad events still works if using document as EventTarget.

I hope these suffices for this to be merged.

src/signature_pad.ts
```diff
@@ -87,12 +128,34 @@ export default class SignaturePad extends EventTarget {
...
+  }
+
+  removeEventListener(
+    type: string,
+    callback: EventListenerOrEventListenerObject | null,
+    options?: boolean | EventListenerOptions,
+  ): void {
+    return this._et.removeEventListener(type, callback, options);
+  }
+
```
UziTech, line 158: It looks like these methods are copied from `SignatureEventTarget` wouldn't it just be easier for `SignaturePad` to extend `SignatureEventTarget`?

src/signature_pad.ts
```diff
@@ -37,13 +37,57 @@ export interface Options extends Partial<PointGroupOptions> {
   velocityFilterWeight?: number;
   backgroundColor?: string;
   throttle?: number;
+  documentAsEventTarget?: boolean;
```
UziTech, line 40: I don't think we should pollute the options object with items used for tests.

src/signature_pad.ts
```diff
@@ -37,13 +37,57 @@ export interface Options extends Partial<PointGroupOptions> {
...
 }
 
 export interface PointGroup extends PointGroupOptions {
   points: BasicPoint[];
 }
 
-export default class SignaturePad extends EventTarget {
+class SignatureEventTarget implements EventTarget {
```
UziTech, line 47: this class should move to it's own file.

UziTech (changes requested):

src/signature_pad.ts
bowozer: I'm happy to remove this in my next commit. However, it wouldn't be able to be tested using document as EventTarget.

src/signature_pad.ts
bowozer: noted. i'm happy to change this in my next commit.

src/signature_pad.ts
UziTech: You should be able to do something like `signpad['_et'] = document` to force it to be a document.

src/signature_pad.ts
bowozer: yes, you're right. Using composition over inheritance here won't get better advantage. There wouldn't be other `EventTarget` behaviour, would it? However when it does need it or this class needs to inherit other class, then perhaps we could implements the `SignatureEventTarget`, instead of extends it.

bowozer: ah yeah, i'll change this in my next commit.

src/signature_pad.ts
bowozer: Good tip, will try it right away.

src/signature_pad.ts
bowozer: hey since you're currently online, do you prefer 'signature-event-target.ts' or 'signature_event_target.ts' ? What's the convention in this project?

tests/signature_pad.test.ts
```diff
@@ -119,22 +119,116 @@ describe('user interactions', () => {
...
+        const event = <CustomEvent>eventDispatched;
+        expect(event.detail).toBe(pointerEvent);
+      });
+    });
+  });
+
+  [{ useDocument: true }, { useDocument: false }].forEach((param) => {
+    describe(`use document as EventTarget=${param.useDocument}.`, () => {
+      beforeEach(() => {
+        signpad['_et'] = document;
```
UziTech, line 208:
```suggestion
        if (param.useDocument) {
          signpad['_et'] = document;
        }
```

UziTech (changes requested):

tests/signature_pad.test.ts
bowozer: ah thanks for reminding me about this. But I think I will implement it differently since testing with EventTarget object has been done at previous test case (above). I will provide with a new commit.

UziTech (approved):

bowozer:
@UziTech Thanks for approving this PR. 

So are we (am i) done here? what's the next step?

UziTech:
Looks good to me we will try to get out a new version with this fix soon.

PR #583, opened 2021-12-16
updated script tag example

jonneroelofs:
Hi,

A made a small change to the readme.

I noticed that the readme has examples (to be specific addEventListener) which assume you are using v4 of the signature pad. 

However the script tag was still including v2. 

Cheers


UziTech (approved):
Thanks! :100:

PR #585, opened 2021-12-21
mouseleave / touchleave events

leviwilson:
In our application, we pop up a modal dialog that has a canvas to capture a signature. If you are drawing, the canvas is appropriately updated but if you drag the signature _off_ of the canvas, the `mouseup` event does _not_ fire, therefor the `strokeEnd` end does not _either_.

This does *not* appear to be an issue with your demo, so I'm unsure of the nuance as to why this is a problem for us, but locally if I did the following:

* do a `canvas.addEventListener('mouseleave', onEndStroke)` on the `beginStroke` event
* on my `onEndStroke` handler, added a `canvas.removeEventListener('mouseleave', onEndStroke)`

Our problem goes away. This PR is to codify that same behavior in the upstream so we only have to then subscribe to `beginStroke` / `endStroke` events (rather than adding our own event handler here).

### DIFF
--- a/docs/js/signature_pad.umd.js
+++ b/docs/js/signature_pad.umd.js
@@ -270,10 +270,12 @@
             document.removeEventListener('pointerup', this._handlePointerEnd);
             this.canvas.removeEventListener('mousedown', this._handleMouseDown);
             this.canvas.removeEventListener('mousemove', this._handleMouseMove);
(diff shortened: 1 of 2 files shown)


leviwilson:
Closing this for a second; I still need to suss out some issues regarding when you leave / come back

PR #591, opened 2022-01-17
fix fast double-tap on iOS 15

marmooo:
On iOS 15, a fast double-tap will trigger an unintended pan event.
I think this is a problem caused by the loupe feature in iOS 15.
It does not occur in iOS 14.
The solution is to always turn off the pan event in the constructor,
instead of switching it in event handling.
I think there is no need to turn the pan event on/off.
Therefore, it is better to stop the pan event in the constructor
so that it is not affected by the browser implementation.

I have checked this patch on the following environments.
- iPhone 8 (iOS 15.1) + Safari    # need fixing
- iPhone 8 (iOS 14.8.1) + Safari  # not need fixing
- MediaPad M3 Lite 10 wp (Android 7.0) + Chrome  # not need fixing
I don't have Windows tablet, so I can't check ms-touch-action,
but theoretically it should work.

https://user-images.githubusercontent.com/28778641/149785412-9f370aaf-d329-4a51-84bd-6bd235f446fe.mp4

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -87,6 +87,10 @@ export default class SignaturePad extends SignatureEventTarget {
 
     this.clear();
 
+    // Disable panning/zooming when touching canvas element
+    this.canvas.style.touchAction = 'none';
+    this.canvas.style.msTouchAction = 'none';
+
     // Enable mouse and touch event handlers
     this.on();
   }
@@ -148,10 +152,6 @@ export default class SignaturePad extends SignatureEventTarget {
   }
 
   public on(): void {
-    // Disable panning/zooming when touching canvas element
-    this.canvas.style.touchAction = 'none';
-    this.canvas.style.msTouchAction = 'none';
-
     const isIOS =
       /Macintosh/.test(navigator.userAgent) && 'ontouchstart' in document;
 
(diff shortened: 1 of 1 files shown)


marmooo:
Another solution is as follows.

```
this.canvas.style.webkitUserSlect = 'none';
```

With this method, you can draw a line, but an obtrusive loupe will appear.
I used the following as a reference for this solution.

- [REGRESSION (iOS 15): Safari shows zoom callout even if -webkit-user-select is none](https://bugs.webkit.org/show_bug.cgi?id=231161)

UziTech:
Are you calling `.off`? If you never call `.off` it shouldn't matter if this gets changed in the `.on` function or before it.

marmooo:
Yes, I have checked on http://szimek.github.io/signature_pad/.
By the way, this URL is 4.0.0, but the result is the same for local 4.0.1.
The following is what is happening in fast double-tap, I think.
This is a guess.

1. [hand] touch start canvas
2. [brower] check touch-action and call js (**some delay**)
3. [js] call `.on` **async** functions and `touch-action:none;`
4. [hand] touch end canvas
5. [js] call `.off` **async** functions and `touch-action:auto;`
6. [hand] touch start canvas
7. [brower] check touch-action and call js (**some delay**)
8. The loupe appears and skip draw line (**occurs problem!**)
9. [js] call `.on` **async** functions and `touch-action:none;`

UziTech:
I don't believe `.off` is ever called in https://szimek.github.io/signature_pad

marmooo:
Sorry, I apologize for my misunderstanding and lack of checking.
I'll take down pull-request.
I don't know what the solution is at this point,
but the problem itself appears to remain, so I will reconsider.

src/signature_pad.ts
```diff
@@ -87,6 +87,8 @@ export default class SignaturePad extends SignatureEventTarget {
 
     this.clear();
 
+    this.canvas.style.userSelect = 'none';
```
UziTech, line 90: This should be in the `on` function with the other style adjustments. Also remove this style in the `off` function.

UziTech (changes requested):

PR #601, opened 2022-02-21
Fix race condition / edge case in _strokeUpdate where this._data can sometimes be []

M1ke:
Replicates https://github.com/szimek/signature_pad/pull/481/ but into v2 branch.

There are other libraries that depend upon older versions of this library and may not want to carry out refactoring to upgrade through 2 major versions to receive this fix.

An example is discussed here https://github.com/agilgur5/react-signature-canvas/pull/68

As the fix is very simple, and maps exactly from v4 back to v2, I am proposing this is merged and a new minor or point release of v2 is cut to include it.

### DIFF
--- a/src/signature_pad.js
+++ b/src/signature_pad.js
@@ -154,6 +154,13 @@ SignaturePad.prototype._strokeBegin = function (event) {
 };
 
 SignaturePad.prototype._strokeUpdate = function (event) {
+  if (this._data.length === 0) {
+      // This can happen if clear() was called while a signature is still in progress,
+      // or if there is a race condition between start/update events.
+      this._strokeBegin(event)
+      return
+  }
+  
   const x = event.clientX;
   const y = event.clientY;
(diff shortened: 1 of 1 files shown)


UziTech:
I think there are better ways to handle https://github.com/agilgur5/react-signature-canvas/pull/68 than to try to maintain old versions of signature_pad. see https://github.com/agilgur5/react-signature-canvas/pull/68#issuecomment-1049518389

M1ke:
I'd agree if there were a range; however this is just one change, and is for what seems to be quite a long standing issue. The benefit of this change is that it requires no changes downstream, and we already know this change fixes the issue in v4, so it should work here. So I guess it depends who has the responsibility, but the change here is definitely the lowest overall amount of work having to be done by people on aggregate.

UziTech:
I don't think we are going to maintain two versions of this library.

PR #615, opened 2022-04-09
docs: add options to `fromDataURL` (and `toDataURL`)

agilgur5:
## Summary

Add two examples to the README for usage of options with `fromDataURL` and `toDataURL`.
- Apply some misc formatting to the README as well

## Details

- some options were added in [v4.0.0](https://github.com/szimek/signature_pad/blob/master/CHANGELOG.md#400) but were not mentioned in the docs
  - I've pointed [multiple](https://github.com/agilgur5/react-signature-canvas/issues/75#issuecomment-1089272704) [users](https://github.com/agilgur5/react-signature-canvas/issues/60#issuecomment-1030713591) to the code for this (instead of the docs) as a result
  - `fromDataURL` had partial options available [earlier](https://github.com/szimek/signature_pad/blob/8145dd499974df25592be7dbd4c583079a9a26c7/src/signature_pad.js#L99), but no examples, and then offsets were added in v4
  - `toDataURL` had options available [earlier](https://github.com/szimek/signature_pad/blob/8145dd499974df25592be7dbd4c583079a9a26c7/src/signature_pad.js#L113) but similarly no examples

- `fromDataURL`: add an example with full `options`
  - add a link to [MDN's `drawImage` docs](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage) for the reference, similar to `toDataURL`'s existing link
- `toDataURL`: add an example with `encoderOptions` on a JPEG

- misc formatting: add new lines after various headings in the README to comply with `markdownlint` etc
  - and after/before code blocks as well
  - also ensure that the language in the code block is immediately after the backticks
    - sometimes there was a space after, but most of the time there wasn't, so just be consistent

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,30 +10,39 @@ It works in all modern desktop and mobile browsers and doesn't depend on any ext
 [Demo](http://szimek.github.io/signature_pad) works in desktop and mobile browsers. You can check out its [source code](https://github.com/szimek/signature_pad/blob/gh-pages/js/app.js) for some tips on how to handle window resize and high DPI screens. You can also find more about the latter in [HTML5 Rocks tutorial](http://www.html5rocks.com/en/tutorials/canvas/hidpi).
 
 ### Other demos
+
 - Erase feature: <https://jsfiddle.net/szimek/jq9cyzuc/>
 - Undo feature: <https://jsfiddle.net/szimek/osenxvjc/>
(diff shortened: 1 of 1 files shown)


UziTech (approved):
Thanks for adding those.

PR #622, opened 2022-06-02
fix(signature_pad.ts): compatible with mobile terminal multi-finger o…

fuaiyi:
Touch events are preferentially triggered to be compatible with multi-finger operations on the mobile


src/signature_pad.ts
```diff
@@ -248,15 +243,18 @@ export default class SignaturePad extends SignatureEventTarget {
...
+      this._strokeMoveUpdate(touch);
+    }
   };
 
   private _handleTouchEnd = (event: TouchEvent): void => {
     const wasCanvasTouched = event.target === this.canvas;
     if (wasCanvasTouched) {
       event.preventDefault();
 
+      this._drawningStroke = false;
```
UziTech, line 257: Do we need to worry about this ending when a touch is still occuring?

For example:

1. First finger touch (`_handleTouchStart` called)
1. Second finger touch (`_handleTouchStart` called)
1. Second finger removed (`_handleTouchEnd` called)
1. First finger moves but nothing happens because `_drawningStroke` is set to false when the second finger was removed.

UziTech:
closing as stale. If you would like to continue working on this feel free to re-open

PR #634, opened 2022-07-17
fix: check for event.cancelable in touch events

jschaf:
Chrome has started warning when canceling an event that cannot be cancelled, like scrolling.
This causes a fair bit of log spam. I know it fires for touchmove and touchend and there's 
reports that it also fires for touchstart.

    IgnoredEventCancel: intervention: Ignored attempt to cancel a touchmove event with
    cancelable=false, for example because scrolling is in progress and cannot be interrupted

References:

- https://github.com/react-grid-layout/react-draggable/issues/553
- https://www.uriports.com/blog/easy-fix-for-intervention-ignored-attempt-to-cancel-a-touchmove-event-with-cancelable-false/
- https://github.com/FL3NKEY/scroll-lock/issues/19
- https://stackoverflow.com/a/53315365/30900

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -236,7 +236,9 @@ export default class SignaturePad extends SignatureEventTarget {
 
   private _handleTouchStart = (event: TouchEvent): void => {
     // Prevent scrolling.
-    event.preventDefault();
+    if (event.cancelable) {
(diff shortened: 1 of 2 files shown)


UziTech:
Can you write a test to make sure this is not broken in future updates?

PR #666, opened 2022-10-18
Adding `import`

ThomasLandauer:


UziTech:
How it is imported will depend on how it is loaded so this won't be valid for everyone.

ThomasLandauer:
I had trouble yesterday figuring this out, so I thought I'd share it.
Which other ways do exist? Maybe you could show them *all*, like in the "Installation" section where you're showing *multiple* ways.

UziTech:
If you add the script tag to the page than it is added as a global so there is no need to import.

ThomasLandauer:
So what about something like this?:
```javascript
// If you don't add it via `<script>`:
import SignaturePad from "signature_pad";
```

UziTech:
Where it says
> This library is provided as UMD (Universal Module Definition) and ES6 module.

We could change it to:
> This library is provided as UMD (Universal Module Definition) and ES6 module with a default export.

ThomasLandauer:
I didn't know the `signature_pad` part of this line - so I had to guess it:

> import SignaturePad from "signature_pad";

This info is what I'm trying to get into README ;-)
Or is this *always* the same as in `yarn add signature_pad` - so that everybody knows it anyway?

UziTech:
Ya that is just the package name. It is always the same to install and import.

ThomasLandauer:
OK, thanks, then let's leave it as it is ;-)

Issue #670, opened 2022-10-29
Bug when scale screen or mobile view


dbannik:
![Screenshot from 2022-10-29 17-12-25](https://user-images.githubusercontent.com/16108701/198833467-904d2f8f-5957-40bc-833f-a76bc19332f2.png)
Next button
```js
const img = signaturePad.toDataURL('image/svg+xml');
signaturePad.fromDataURL(img);
````
![Screenshot from 2022-10-29 17-12-30](https://user-images.githubusercontent.com/16108701/198833466-3a43967f-52ec-4764-87e8-be6a67a9a445.png)

preview https://jsfiddle.net/dbannik/0hvs9ezo/15/

**Comment by UziTech** [comment]
I can't reproduce this on a laptop with 1.25 DPI

https://user-images.githubusercontent.com/97994/198900541-d34fe678-ed8d-4b2f-9866-e5220cfda9ad.mp4

Instead of doing

```js
const img = signaturePad.toDataURL('image/svg+xml');
signaturePad.fromDataURL(img);
```

Would it work to do 

```js
const data = signaturePad.toData();
signaturePad.fromData(data);
```

UziTech:
@ocarreterom your issue is that your canvas size does not match your canvas css size. The canvas size defaults to 300px X 150px. When you set the css size to 100% you are scaling the canvas without changing it's size. If you want it to be a percent width you should resize the canvas when the css size changes. https://github.com/szimek/signature_pad/blob/master/docs/js/app.js#L19

marmooo:
It's a similar topic, I think 4.1.2 (#673) is broken when DPI != 1. Simple example is below.
```
<html>
  <head>
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.2/dist/css/bootstrap.min.css" rel="stylesheet">
    <style> canvas { width:300px; height:300px; border:1px solid #000; } </style>
  </head>
  <body>
    <main class="container">
      <div style="height:200px; background:black;">This causes a bug.</div>
      <div id="dpr"></div>
      <canvas id="canvas" width="300" height="300"></canvas>
    </main>
    <script src="https://cdn.jsdelivr.net/npm/signature_pad@4.1.3/dist/signature_pad.umd.min.js"></script>
    <script>
      const dpr = Math.max(window.devicePixelRatio || 1, 1);
      document.getElementById("dpr").textContent = dpr;
[...]

PR #671, opened 2022-10-29
Fixed scale SVG export when dynamic DPI

dbannik:
resolve https://github.com/szimek/signature_pad/issues/670


UziTech:
This would be a pretty big breaking change. I don't think we want to change the signature for `toDataURL` for pngs and jpgs. We want to keep it the same as the [`canvas.toDataURL()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL) method.

PR #672, opened 2022-10-31
fix: Fix empty `toDataURL()` in TypeScript

indianakernick:
The recent change to the `toDataURL` function made the `type` parameter required. It was previously optional. This is a breaking change and most likely an innocent mistake.

Additionally, the type of the `encoderOptions` parameter for the second overload was corrected. The type of `type` could be `string` and its value could be `'image/svg+xml'`. So `ToSVGOptions` should be accepted by the `encoderOptions` parameter. This is not technically a breaking change but it's something I noticed while I was in the area.

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -146,7 +146,7 @@ export default class SignaturePad extends SignatureEventTarget {
     type: 'image/svg+xml',
     encoderOptions?: ToSVGOptions,
   ): string;
-  public toDataURL(type?: string, encoderOptions?: number | ToSVGOptions): string;
+  public toDataURL(type?: string, encoderOptions?: number): string;
   public toDataURL(
     type = 'image/png',
     encoderOptions?: number | ToSVGOptions | undefined,
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -146,7 +146,7 @@ export default class SignaturePad extends SignatureEventTarget {
     type: 'image/svg+xml',
     encoderOptions?: ToSVGOptions,
   ): string;
-  public toDataURL(type?: string, encoderOptions?: number): string;
+  public toDataURL(type?: string, encoderOptions?: number | ToSVGOptions): string;
   public toDataURL(
     type = 'image/png',
     encoderOptions?: number | ToSVGOptions | undefined,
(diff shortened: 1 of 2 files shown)


UziTech:
> The type of `type` could be string and its value could be `'image/svg+xml'`. So `ToSVGOptions` should be accepted by the encoderOptions parameter.

The whole point of the overload is so the `encoderOptions` parameter can be `ToSvgOptions` only when the `type` is `'image/svg+xml'`

UziTech (changes requested):
Can you add a test for `toDataURL` without a type parameter so this doesn't break again.

indianakernick:
> Can you add a test for `toDataURL` without a type parameter so this doesn't break again?

Added.

> The whole point of the overload is so the `encoderOptions` parameter can be `ToSvgOptions` only when the type is `'image/svg+xml'`

Yeah. I understand the intent. When `type` is `'image/svg+xml'`, `encoderOptions` must be `ToSvgOptions`. When it is not `'image/svg+xml'`, it must be `number`. It's a bit difficult to express that second part in TypeScript currently. There's no way to say "any `string` except for `'image/svg+xml'`". Because of this, the overload set doesn't really work. For example, this compiles: `toDataURL('image/svg+xml', 1)`. In this case, the second overload is chosen.

As it is defined with and without the changes in this PR, `toDataURL` doesn't fully enforce correct usage at compile-time. However, the first overload still expresses intent. It still acts as a form of documentation. When I type `toDataURL('image/`, the IDE suggests that I use `'image/svg+xml'` and pass `ToSvgOptions` as the second argument. There are a few possibilities for improving this.

 1. Throw an exception instead of ignoring `encorderOptions` when the runtime type check fails.
 2. Remove the SVG overload from `toDataURL` and instead create a dedicated function that calls `toSVG` and converts it to a data URL. This is a breaking change.
 3. Instead of using `string`, use a union of all image mime types. So `type ImageMimeType = 'image/png' | 'image/svg+xml' ...`. Then the first overload could use `'image/svg+xml'` and the second overload could be `Exclude<ImageMimeType, 'image/svg+xml'>`. This is also a breaking change. Also, maintaining a complete list of image mime types would be a bit of a pain.

Unless you feel like releasing version 5.0.0 with some breaking changes, it seems like the available options are limited.

UziTech:
I don't care if they put a number for SVG. It will be handled the same as setting a number for PNG It does nothing. But I don't want ToSVGOption except for SVG.

UziTech:
Thanks for bringing this to my attention! :100:

PR #691, opened 2023-01-28
Close #690, add public `.scale()` to allow supported scaling of signature

### LINKED ISSUE #690: Add `signaturePad.scale(x, y)`
Scaling the signature lets a developer adjust its size according to the canvas it's in. So if the canvas shrinks or grows, the developer can do the same to the image. I have an example of using such a function at https://jsfiddle.net/j2Lurpd5/1/. I've written a comment showing the code at https://github.com/szimek/signature_pad/issues/32#issuecomment-1406826837. I'll make a PR with the code for a `.scale()` method.

Feature

There is no way for a developer to scale the signature without counting on the current internal structure of a SignaturePad. That is, if the interface changes, the developer would need to change their code as well. If SignaturePad handled it, this wouldn't be a concern.

Allow access to a `.scale(x, y)` method so that the developer doesn't have to touch the internal structure of a SignaturePad.

I'm not sure if this is really appropriate for this library. From other comments I've seen, I get the impression that this library is trying to keep its scope very focused. If this change doesn't fit that scope, I understand. If this isn't the right place to solve this, does anyone know of a non-React wrapper library that does this?

plocket:
Allow access to a `signaturePad.scale(x, y)` method so that the developer doesn't have to touch the internal structure of a SignaturePad to resize their signatures for a new canvas size. Closes #690 (see for more details).

I'm not sure why yarn updated. I can get rid of those changes if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,9 @@ signaturePad.fromData(data);
 // Draws signature image from an array of point groups, without clearing your existing image (clear defaults to true if not provided)
 signaturePad.fromData(data, { clear: false });
 
+// Scales the size of the image by multiplying by the given scale. To keep the aspect ratio the same, use the same number for both x and y.
+signaturePad.scale(xScale, yScale);
+
 // Clears the canvas
 signaturePad.clear();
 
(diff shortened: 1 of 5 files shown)


UziTech:
The biggest problem with scaling like this is it doesn't scale the line width so it won't actually be a perfect scaling.

Issue #421, opened 2019-02-16
Erase and Undo don't work well together


fesnow:
I wanted to use an erase feature and an undo feature on the same canvas, I added those 2 features with the same ways as the official demos.
```javascript
document.getElementById('erase').addEventListener('click', function () {
  var ctx = canvas.getContext('2d');
  ctx.globalCompositeOperation = 'destination-out';
});

document.getElementById('undo').addEventListener('click', function () {
  var data = signaturePad.toData();
  if (data) {
    data.pop(); // remove the last dot or line
    signaturePad.fromData(data);
  }
});
```
This is the code on JSFiddle https://jsfiddle.net/fesnow/4eupn3j7/3/

## 1. Using undo feature after erasing makes all the lines invisible
![screencapture](https://user-images.githubusercontent.com/1832504/52896349-f24e1580-31c6-11e9-9e46-a19702fd6832.gif)

## 2. It looks working well with turning on draw feature before using undo feature
![screencapture](https://user-images.githubusercontent.com/1832504/52896405-b9627080-31c7-11e9-85c5-426fbda6b88f.gif)
Turning on draw feature sets `ctx.globalCompositeOperation = 'source-over'`

## 3. Even with 2. way, if I use erase feature multi times, erased lines become like drawn lines
![screencapture](https://user-images.githubusercontent.com/1832504/52896472-d9defa80-31c8-11e9-9062-11f882491232.gif)

I was using v2.3.2 on my local, it happens with v3.0.0-beta.3 as well (the code on JSFiddle), I checked with Chrome, Safari, and Firefox.

Is there a way to solve this problem? And this is why the demos are separated?
I think it's not a rare case,  I hope it will help many people.
Thank you.

tejacoder:
same case , please fix this, thx

JoeLeung32:
My solution for v3.0.0-beta.3

In `signature_pad.umd.js`

```
SignaturePad.prototype._strokeBegin = function (event) {
    var ctx = event.target.getContext('2d');    // Fix undo problem
    var newPointGroup = {
        color: this.penColor,
        dotSize: this.dotSize,  // Enhance undo
        minWidth: this.minWidth,    // Enhance undo
        maxWidth: this.maxWidth,    // Enhance undo
        compositeOperation: ctx.globalCompositeOperation,   // Fix undo problem
        points: []
    };
    ...
};
```

```
SignaturePad.prototype._drawDot = function (_a) {
    var color = _a.color, point = _a.point;
[...]

PR #725, opened 2023-06-25
fix(eraser): solves issues with eraser

NLueg:
Hi there! 

I struggled a lot with implementing an `undo` and `erase` feature for my canvas.
After research, I found issue #421 which already provides a well-working version.

This PR includes the suggested changes which are provided by @JoeLeung32 (btw. thank you so much!)

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -36,6 +36,7 @@ export interface PointGroupOptions {
   maxWidth: number;
   penColor: string;
   velocityFilterWeight: number;
+  compositeOperation: GlobalCompositeOperation;
 }
 
 export interface Options extends Partial<PointGroupOptions> {
@@ -56,6 +57,7 @@ export default class SignaturePad extends SignatureEventTarget {
   public penColor: string;
   public minDistance: number;
   public velocityFilterWeight: number;
+  public compositeOperation: GlobalCompositeOperation;
   public backgroundColor: string;
   public throttle: number;
 
@@ -83,6 +85,7 @@ export default class SignaturePad extends SignatureEventTarget {
     this.dotSize = options.dotSize || 0;
     this.penColor = options.penColor || 'black';
     this.backgroundColor = options.backgroundColor || 'rgba(0,0,0,0)';
+    this.compositeOperation = options.compositeOperation || 'source-over';
 
     this._strokeMoveUpdate = this.throttle
       ? throttle(SignaturePad.prototype._strokeUpdate, this.throttle)
@@ -314,7 +317,7 @@ export default class SignaturePad extends SignatureEventTarget {
     }
   };
 
-  private _getPointGroupOptions(group?: PointGroup) {
+  private _getPointGroupOptions(group?: PointGroup): PointGroupOptions {
     return {
       penColor: group && 'penColor' in group ? group.penColor : this.penColor,
       dotSize: group && 'dotSize' in group ? group.dotSize : this.dotSize,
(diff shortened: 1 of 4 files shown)


UziTech:
This library is meant to be for signatures so an erase feature doesn't really make sense. If you want to create a drawing_pad library that does this that would make more sense but I don't think an erase feature will be added to this library.

NLueg:
I know, that it's not the main focus of this plugin. But actually, you have a Stackblitz link for the ´Erase feature` at the top of the README.
Currently, this feature doesn't work although you advertise it there. We can close this PR if this is not intended but maybe you should make this clear in the README and remove the Stackblitz link.

After struggling a lot I just found the issue I mentioned and the fix from more than 4 years ago worked pretty well.

UziTech:
@szimek what do you think about this feature? We would just need to add `compositeOperation` to `pointGroupOptions`, or we could abstract it with something like `isErase`?

src/signature_pad.ts
```diff
@@ -46,6 +46,8 @@ export interface Options extends Partial<PointGroupOptions> {
 
 export interface PointGroup extends PointGroupOptions {
   points: BasicPoint[];
+  color: string;
+  compositeOperation: GlobalCompositeOperation;
```
UziTech, line 50: penColor is already part of PointGroupOptions. compositeOperation should be part of PointGroupOptions.

src/signature_pad.ts
```diff
@@ -333,8 +335,14 @@ export default class SignaturePad extends SignatureEventTarget {
 
     const pointGroupOptions = this._getPointGroupOptions();
 
+    const ctx = (event.target as HTMLCanvasElement).getContext('2d');
     const newPointGroup: PointGroup = {
       ...pointGroupOptions,
+      color: this.penColor,
+      dotSize: this.dotSize,
+      minWidth: this.minWidth,
+      maxWidth: this.maxWidth,
```
UziTech, line 344: These should already be in pointGroupOptions

src/signature_pad.ts
```diff
@@ -582,6 +590,12 @@ export default class SignaturePad extends SignatureEventTarget {
             basicPoint.time,
           );
 
+          this.penColor = color;
+          this.minWidth = minWidth;
+          this.maxWidth = maxWidth;
```
UziTech, line 595: Why are these set?

UziTech (changes requested):

NLueg:
@UziTech thanks for the fast review! Actually, I don't thought about the changes proposed in the issue. I refactored the code as you requested and verified it in my fork.
It still works perfectly as expected and is much less hacky. Thank you!

src/signature_pad.ts
NLueg: I just done copy + paste without thinking 😄

src/signature_pad.ts
NLueg: You're right! This is completely obsolet.
I just done copy + pasted without thinking 😄

UziTech (changes requested):
Thanks for simplifying it! Can you write a new test with erase and fix current tests?

NLueg:
> Thanks for simplifying it! Can you write a new test with erase and fix current tests?

I added a test that checks whether the set `compositeOperation` is applied to the canvas context. Do you think that is enough? Or what test do you want to have?

UziTech (approved):
LGTM

yserlong:
Hello,
since this change my project won't build anymore. It throw an error telling it can't find the type : GlobalCompositeOperation.
Maybe it's a compatibility problem with the typescript version or something ?

UziTech:
Looks like this may require Typescript 4.6 or higher

root5427:
Hello. Looks like this PR broke the erase feature that was working fine by setting up the `globalCompositeOperation` of the canvas context. I applied the new `compositeOperation` of `SignaturePad` but it is not working as expected. Check this fiddle to make sure it doesn't work properly: https://jsfiddle.net/jq9cyzuc/.

I would appreciate any suggestions to get this all working, because I may not have figured out how to work with `compositeOperation` due to my poor knowledge of the Canvas API.

PR #740, opened 2023-10-20
Update signature_pad.ts

simongeau:
This allow the canvas width and height attribute to be different from the canvas CSS style dimensions allowing it to be responsive;

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -450,9 +450,13 @@ export default class SignaturePad extends SignatureEventTarget {
   private _createPoint(x: number, y: number, pressure: number): Point {
     const rect = this.canvas.getBoundingClientRect();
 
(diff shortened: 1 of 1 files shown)


UziTech:
Duplicate of #446 

See https://github.com/szimek/signature_pad/pull/446#issuecomment-1059888995 on why this is a bad idea

simongeau:
Not sure on the argument; scaling and ratio are 2 different things and legal validity of the signature should be an app developper concern not library concern.

Anyway, I will work with my forked code!

UziTech:
> legal validity of the signature should be an app developer concern not library concern.

I totally agree, unfortunately most app developers won't even know to look into it until it is too late. We want to make the easy path the way least developers can screw up.

PR #743, opened 2023-11-14
Added Apple Pencil only mode

ImOlli:
This adds a new option to the signature pad to only use it with an Apple Pencil. This is useful if used on an iPad so you don't accidentally write with your palm or fingers wile holding the pen.

The Flag `touch.touchType` only exists on iOS and is either "stylus" when it's used by a pen or "direct" when the event is triggered by finger touch or something.

### DIFF
--- a/docs/js/signature_pad.umd.js
+++ b/docs/js/signature_pad.umd.js
@@ -165,6 +165,7 @@
             this._data = [];
             this._lastVelocity = 0;
             this._lastWidth = 0;
+            this._isIOS = /Macintosh/.test(navigator.userAgent);
             this._handleMouseDown = (event) => {
                 if (event.buttons === 1) {
                     this._drawningStroke = true;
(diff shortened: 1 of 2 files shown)


UziTech:
If  it is something that will never be standardized we probably shouldn't add it to the library. If it will be standardized we should make it a generic option. We don't want someone to have to set `onlyAllowApplePencil` when they want to allow other styluses.

ImOlli:
@UziTech I understand your concerns. 
The pointer event has this standardized [pointer type](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType) and it works for other devices also. Due to the problematic on iOS devices with the pointer event we use touch event here for iOS. The touch event only has the flag on iOS as I described earlier.

Would it be fine for you if I rebuild this to a generic version where the event is filtered on pointer event and touch event? And the option would be something like `onlyAllowPens`


I need this feature for a project I'm working on. 
If you don't think we should add this to the library. Could we implement a way to cancel the touch event in the beginStroke listeners or something on the signature_pad so I can implement it myself?

UziTech:
I actually like the idea of the beginStroke event being cancelable better.

UziTech:
I created #744 to allow canceling beginStroke

ImOlli:
That sounds good for me. What about this PR should I still build the generic version or should we close this?

PR #752, opened 2023-12-26
feat: Adjust drawing line and mouse position according to CSS zoom level

buYoung:
Added ability to match mouse position to line drawn for zoom level in CSS

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -65,6 +65,7 @@ export default class SignaturePad extends SignatureEventTarget {
   public compositeOperation: GlobalCompositeOperation;
   public backgroundColor: string;
   public throttle: number;
+  public zoom: number;
 
   // Private stuff
   /* tslint:disable: variable-name */
@@ -94,6 +95,7 @@ export default class SignaturePad extends SignatureEventTarget {
     this.penColor = options.penColor || 'black';
     this.backgroundColor = options.backgroundColor || 'rgba(0,0,0,0)';
     this.compositeOperation = options.compositeOperation || 'source-over';
+    this.zoom = 1.0;
 
     this._strokeMoveUpdate = this.throttle
       ? throttle(SignaturePad.prototype._strokeUpdate, this.throttle)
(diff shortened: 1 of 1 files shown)


UziTech:
I don't get it, why is this needed? why not just set css zoom level to 1?

buYoung:
> I don't get it, why is this needed? why not just set css zoom level to 1?

Thanks for your comment. This PR tackles issues with mouse accuracy on the canvas when CSS zoom is adjusted, affecting both zoom levels below and above 1. It ensures precise interaction for all, enhancing overall canvas usability

UziTech:
but the developer will have to set the zoom level, it is not set automatically to the zoom level for css. How about instead of setting the zoom level to equal css zoom level the developer sets the css zoom level equal to the signature_pad zoom level? which is 1

buYoung:
> but the developer will have to set the zoom level, it is not set automatically to the zoom level for css. How about instead of setting the zoom level to equal css zoom level the developer sets the css zoom level equal to the signature_pad zoom level? which is 1

Unfortunately, based on my tests, applying CSS zoom to the root element affects the alignment between the mouse position and the drawn line on the canvas, even when adjustments are made at lower levels.

UziTech:
> Unfortunately, based on my tests, applying CSS zoom to the root element affects the alignment between the mouse position

Right so wouldn't the correct fix be **not** applying CSS zoom instead of trying to fix it with this method?

buYoung:
> > Unfortunately, based on my tests, applying CSS zoom to the root element affects the alignment between the mouse position
> 
> Right so wouldn't the correct fix be **not** applying CSS zoom instead of trying to fix it with this method?

Thanks for your suggestion. However, not using CSS zoom might not solve our problem. Here's why: When we create web pages, some people might want to see things bigger, like those who prefer larger text. Using CSS zoom does make everything bigger, but it can cause a mismatch between where the mouse is pointing and where lines are actually drawn on the canvas. The feature we've developed helps to fix this issue. It makes sure that even if the screen is zoomed in, the mouse position is still accurate. That's why I think this method is a better solution!

UziTech:
Can you create a jsfiddle with this fix and show that this will actually work when zooming in? Since the developer will need to set the correct zoom I am very skeptical that this will work on all devices.

UziTech:
Going to close this as stale. If you would like to continue working on it feel free to reopen it

PR #761, opened 2024-03-09
Add canvasContextOptions API for use by the getContext

### LINKED ISSUE #760: Include the willReadFrequently in getContext
Hello Team,

We are trying to fix an issue with signature_pad in our project where the initial value was not set properly. After a bit of research we made it fixed by using `getImageData`, and `putImageData`. We also noticed that calling the `getContext` with `willReadFrequently` enabled on our side would raise a warning in the browser. Further investigation led us to the conclusion that we also need to modify the very first call to `getContext`, which is done by signature_pad. After that, the warning disappeared.

Link to our project https://github.com/Megabit/Blazorise/pull/5349

---

If you agree to the change, we can create a PR to include that change in here.

stsrki:
Closes #760


stsrki:
I added the `willReadFrequently` option to be passed to the `canvas.getContext`. Since the name `willReadFrequently` might be too generic, I propose renaming it to

- `willReadCanvasFrequently`, or
- `willReadContextFrequently`, or
- `willReadCanvasContextFrequently`

UziTech:
I think it would be better to have a contextOptions instead of each option. That way when Chrome adds more options we won't need to change anything.

stsrki:
> I think it would be better to have a canvasContextOptions instead of each option. That way when Chrome adds more options we won't need to change anything.

Makes sense. I will make the changes.

stsrki:
I have update the API and add canvasContextOptions.

package-lock.json
UziTech, line 1: Delete this file and use yarn

README.md
```diff
@@ -105,6 +105,8 @@ signaturePad.on();
 <dd>(string) Color used to draw the lines. Can be any color format accepted by <code>context.fillStyle</code>. Defaults to <code>"black"</code>.</dd>
 <dt>velocityFilterWeight</dt>
 <dd>(float) Weight used to modify new velocity based on the previous velocity. Defaults to <code>0.7</code>.</dd>
+<dt>canvasContextOptions</dt>
+<dd>(CanvasRenderingContext2DSettings) part of the Canvas API, provides the 2D rendering context for the drawing surface of a <code>canvas</code> element. It is used for drawing shapes, text, images, and other objects.</dd>
```
UziTech, line 109: can we link to MDN or something to show which options are available?

UziTech (changes requested):

README.md
stsrki: Done.

UziTech (approved):
Thanks 💯

stsrki:
Thank you for merging this feature. Much appreciated 🙏

PR #763, opened 2024-03-20
docs: Add redo functionality in demo site for signature path restore when undo path.

mdabbas-cse:


UziTech:
seems like `undoData` should be cleared if a mark is made. If the user makes a few marks, then clicks undo, then makes another mark, the redo button should not do anything.

PR #770, opened 2024-05-02
added toBlob method

qxsch:
added toBlob method


UziTech:
You pass the canvas into signature_pad. Why not just call toBlob on that instead of having to do so through the signature_pad object?

This seems like unnecessary maintenance required by signature_pad if the API ever changes.

qxsch:
Ok good point, That‘s how i do it now.

I was seeing, that there is a toDataURL method and was thinking, no toBlob has been forgotten.

PR #771, opened 2024-05-06
feat: highlight

### LINKED ISSUE #726: Pen highlight for drawing on images
Requesting a feature...

When drawing over an image, the pen colour can easily blend in and get lost. In the past I've solved this by adding a 'highlight' colour, which is different from the pen. The contrast allows the new markings to standout better. The way I did it previously was for the 'highlight' colour to be drawn first, and be a number of pixels wider than the pen colour.

Would be cool if this library had something like that, especially since you've got colouring over images in the ReadMe.

UziTech:
Add new options for highlighting strokes

new options:
- highlightColor: default to blank to remove highlight
- highlightSize: default to `1` to have 1 pixel of highlight on each side of the stroke

signature

fixes #726

TODO:
- [ ] update docs
- [ ] need to draw entire stroke on every update


849599470:
> Thanks for the suggestion but that just applies the drop-shadow to the canvas element not the drawing on the canvas.
> 
> image

When initializing Signature, set the background color to transparent.
```js
new Signature(canvasDom, {
    // other options
    backgroundColor: "rgba(0,0,0,0)",
})
```

PR #772, opened 2024-05-18
Use fallback values when options object contains explicit `undefined` values

cyevgeniy:
This PR allows to handle explicit `undefined` values in   `throttle`, `minDistance` or `canvasContextOptions` fields.

Reproduction link: https://codepen.io/username1334/pen/mdYVWVN

In the reproduction above, signature pad doesn't work because `minDistance` is `undefined`.


UziTech (approved):
Thanks 💯

PR #785, opened 2024-08-23
fix: prevent division by zero in bezier

samvimes01:
In some cases Bezier curve calculations could fail due to division by zero.
(for example when signature was generated by older version of the lib or another lib with similar approach to saving canvas data like this one http://keith-wood.name/signature.html)

Screenshot 2024-08-23 at 16 29 42

This PR fixes the issue by checking for zero before the division.


UziTech (changes requested):
Thanks! Can you add a test?

UziTech (approved):

Issue #787, opened 2024-08-29
Drawing with two fingers jumps between points. Can't disable multitouch


piotrpoliwoda:
**Do you want to request a feature or report a bug?**
Report a bug

**What is the current behavior?**
When drawing with two fingers the touch points jump around both fingers.

**Replication steps:**
1. Go to the sample site: https://szimek.github.io/signature_pad/
2. Use two fingers to draw a line.

**Result:**
The signature creates a connecting line between two fingers.

See result screenshot:
![image](https://github.com/user-attachments/assets/b3325477-0f17-47af-8104-3ca1c38f774b)

Do this:
![image](https://github.com/user-attachments/assets/e11ddddb-9891-4622-af37-99a55e357091)

**What is the expected behavior?**
The expected behaviour is that multitouch should be disabled. Looking through the code, it is theoretically trying to prevent a double touch but it doesn't really work on a mobile device. The lines are created and can make a big black square.

**Which versions of SignaturePad, and which browser / device are affected by this issue? Did this work in previous versions of SignaturePad?**

Version: 5.0.3
Browser: All of them, iOS safari and Chrome as well as Android Chrome.

**Some debugging information:**
- On iOS in Safari, the mobile web browser event is being caught by the mousePointer click even though the `pointerType: "touch"`
- When touching the screen, the canvas pointerDown event does not get any of the `event.touches` nor `event.targetTouches` they are undefined 
![image](https://github.com/user-attachments/assets/a9e01652-7a6e-4d56-9f25-4a4c1545fe95)
- There is no easy way to catch if the user is already touching the screen as the drawingStroke turns to false when holding one finger down and pressing down with another finger.
![image](https://github.com/user-attachments/assets/bb362a88-abef-4edf-b3d4-1291acb1b11b)
- See full event data.
![image](https://github.com/user-attachments/assets/c5cde028-357f-4faa-9af2-ebb8c531ea49)

UziTech:
If you would like to create a PR that would be awesome 😁👍

wokayme:
@UziTech I think this should address the problem:
https://github.com/szimek/signature_pad/pull/788
can I ask you for CR and help with testing :) ?

UziTech:
For testing you should need to update the tests with the new property and create a test that has both isPrimary true and false touches.

https://github.com/szimek/signature_pad/blob/267935b1d147367225a9a8c0a216973111bf6e65/tests/signature_pad.test.ts#L266

wokayme:
[...]

PR #788, opened 2024-08-30
fix: prevent multiply pointers drawing

wokayme:
Fix following issue:
https://github.com/szimek/signature_pad/issues/787

Using native api for pointerEvents:
https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/isPrimary

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -376,7 +376,11 @@ export default class SignaturePad extends SignatureEventTarget {
   };
 
   private _handlePointerDown = (event: PointerEvent): void => {
-    if (!this._isLeftButtonPressed(event) || this._drawingStroke) {
+    if (
+      !event.isPrimary ||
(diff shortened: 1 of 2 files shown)


UziTech (changes requested):
Add a test that uses multi touch

src/signature_pad.ts
```diff
@@ -386,7 +390,11 @@ export default class SignaturePad extends SignatureEventTarget {
   };
 
   private _handlePointerMove = (event: PointerEvent): void => {
-    if (!this._isLeftButtonPressed(event, true) || !this._drawingStroke) {
+    if (
+      !event.isPrimary ||
```
UziTech, line 394: I think we should just return if !event.isPrimary adding it to this if will end the stroke. We should also return from _handlePointerUp if !event.isPrimary.

UziTech (changes requested):

PR #815, opened 2025-03-06
Add conditional exports to package.json

lukescott:
I'm in the process of updating to Node 22 and I ran into an issue with the default export not working:

```
src/signature.tsx:25:18 - error TS2351: This expression is not constructable.
  Type 'typeof import("./node_modules/signature_pad/dist/types/signature_pad")' has no construct signatures.

25   this.pad = new SignaturePad(this.canvas, {
```

This PR fixes compatibility with TypeScript's `"nodenext"` module resolution. The `"module"` property is non-standard and not supported by node. The official way is "Conditional Exports":

https://nodejs.org/docs/latest-v22.x/api/packages.html#conditional-exports

This is needed for packages that support CommonJS and ESM where the `"type"` is omitted or set to `"commonjs"`.

(Note: "require" could be omitted, but I was more explicit including both "require" and "default")

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "exports": {
     "import": "./dist/signature_pad.js",
     "require": "./dist/signature_pad.umd.js",
-    "default": "./dist/signature_pad.umd.js",
+    "default": "./dist/signature_pad.umd.js"
   },
   "scripts": {
     "build": "yarn run lint && yarn run clean && rollup --config && yarn run emit-types && yarn run update-docs",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,11 @@
   "main": "dist/signature_pad.umd.js",
   "module": "dist/signature_pad.js",
   "types": "dist/types/signature_pad.d.ts",
+  "exports": {
+    "import": "./dist/signature_pad.js",
+    "require": "./dist/signature_pad.umd.js",
+    "default": "./dist/signature_pad.umd.js",
+  },
   "scripts": {
     "build": "yarn run lint && yarn run clean && rollup --config && yarn run emit-types && yarn run update-docs",
     "clean": "yarn run del dist",


package.json
```diff
@@ -13,6 +13,11 @@
   "main": "dist/signature_pad.umd.js",
   "module": "dist/signature_pad.js",
   "types": "dist/types/signature_pad.d.ts",
+  "exports": {
+    "import": "./dist/signature_pad.js",
+    "require": "./dist/signature_pad.umd.js",
+    "default": "./dist/signature_pad.umd.js",
```
UziTech, line 19: Looks like it has a problem with the trailing comma I this line

UziTech (changes requested):

package.json
lukescott, line 19:
```suggestion
    "default": "./dist/signature_pad.umd.js"
```

lukescott: Fixed

UziTech (approved):
Thanks! 💯

BoussonKarel:
@UziTech tis change has broken my types, because the "exports" field you added has to "types" property:
```json
  "exports": {
    "import": "./dist/signature_pad.js",
    "require": "./dist/signature_pad.umd.js",
    "default": "./dist/signature_pad.umd.js",
    "types": "./dist/types/signature_pad.d.ts"
  },
  ```

**Comment by niklaas** [comment]
For posterity, this could be considered a breaking change, see https://nodejs.org/docs/latest-v22.x/api/packages.html#package-entry-points:

> Existing packages introducing the ["exports"](https://nodejs.org/docs/latest-v22.x/api/packages.html#exports) field will prevent consumers of the package from using any entry points that are not defined, including the [package.json](https://nodejs.org/docs/latest-v22.x/api/packages.html#nodejs-packagejson-field-definitions) (e.g. require('your-package/package.json')). This will likely be a breaking change.
> 
> To make the introduction of ["exports"](https://nodejs.org/docs/latest-v22.x/api/packages.html#exports) non-breaking, ensure that every previously supported entry point is exported. (...)

Our builds started failing and it took me a bit to figure out what's going on. In parts of our codebase we depended on `signature_pad` like this

```js
require('signature_pad/dist/signature_pad.umd')
```

or

```js
import SignaturePad from 'signature_pad/dist/signature_pad.umd'
```

which threw errors such as

```
  ERROR in ./<redacted>.js 142:0-47
  Module not found: Error: Package path ./dist/signature_pad.umd is not exported from package /<redacted>/node_modules/signature_pad (see exports field in /<redacted>/node_
modules/signature_pad/package.json)
```

This can be fixed by changing the above to

```js
require('signature_pad')
```

and

```js
import SignaturePad from 'signature_pad'
```

respectively.

BoussonKarel:
I think niklaas' point is not the fact that we should not have added this, but that the semver was not respected.

v5 should have the old paths listed undert he `exports` key for backwards compatibility [as specified in the node docs](https://nodejs.org/docs/latest-v22.x/api/packages.html#package-entry-points:~:text=To%20make%20the%20introduction%20of%20%22exports%22%20non%2Dbreaking).
```json
  "exports": {
    "import": "./dist/signature_pad.js",
    "require": "./dist/signature_pad.umd.js",
    "default": "./dist/signature_pad.umd.js",
    "./dist/signature_pad.js": "./dist/signature_pad.js",
    "./dist/signature_pad.min.js": "./dist/signature_pad.min.js",
    "./dist/signature_pad.umd.js": "./dist/signature_pad.umd.js",
    "./dist/signature_pad.umd.min.js": "./dist/signature_pad.umd.min.js",

  },
```

And then in 6.0.0 remove it

lukescott:
The intention was to preserve backwards compatibility, and not knowing that import/require/defaults covered more than what "main" and (non-standard) "module" covered. If adding `./dist/` to the exports fixes the issue, a PR can certainly be opened and merged in to fix the breaking change.

PR #853, opened 2025-11-14
fix: handle pointercancel/touchcancel to end stroke correctly

### LINKED ISSUE #852: Bug: pointercancel / touchcancel prevents further strokes
### Description
When a stroke is interrupted by a pen + finger combination or by a system gesture, the browser fires a `pointercancel` or `touchcancel` event.
SignaturePad currently does not handle these events, which causes the internal state to get stuck:
- `_drawingStroke` remains true
- `_strokePointerId` never resets
- new strokes cannot start afterwards

The pad becomes unresponsive until manually reset.

### Expected behavior
When `pointercancel` or `touchcancel` occurs, `SignaturePad` should:
- properly end the current stroke via `_strokeEnd(..., false)`
- reset internal state (`_drawingStroke`, `_strokePointerId`)
- remove temporary move/up listeners
- dispatch `endStroke`
- allow new strokes to begin immediately

### Steps to reproduce
1. Use `SignaturePad` on a touch device with pen support (e.g., Android tablet).
3. Start a stroke using the pen.
5. touch the canvas at the same time as drawing with the pen. → Browser emits `pointercancel` and `touchcancel` .
8. Try drawing again. → Drawing no longer works.

### Root cause
`pointercancel` and `touchcancel` are not handled in the library.
Therefore `_strokeEnd` is never called when a stroke is aborted by the system.

### Proposed fix
```
addEventListener('pointercancel', this._handlePointerCancel, { passive: false });
addEventListener('touchcancel', this._handleTouchCancel, { passive: false });
```

```
private _handlePointerCancel(event: PointerEvent): void {
  if (!this._allowPointerId(event)) return;
  if (event.cancelable) event.preventDefault();
  this._strokeEnd(this._pointerEventToSignatureEvent(event), false);
}

private _handleTouchCancel(event: TouchEvent): void {
  if (event.cancelable) event.preventDefault();
  this._strokeEnd(this._touchEventToSignatureEvent(event), false);
}

```

```
removeEventListener('pointercancel', this._handlePointerCancel);
removeEventListener('touchcancel', this._handleTouchCancel);
```

### Further Readings

- https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event
- https://developer.mozilla.org/en-US/docs/Web/API/Element/touchcancel_event

maat-code:
closes #852

### DIFF
--- a/src/signature_pad.ts
+++ b/src/signature_pad.ts
@@ -125,6 +125,8 @@ export default class SignaturePad extends SignatureEventTarget {
     this._handlePointerDown = this._handlePointerDown.bind(this);
     this._handlePointerMove = this._handlePointerMove.bind(this);
     this._handlePointerUp = this._handlePointerUp.bind(this);
+    this._handlePointerCancel = this._handlePointerCancel.bind(this);
+    this._handleTouchCancel = this._handleTouchCancel.bind(this);
 
     this._ctx = canvas.getContext(
       '2d',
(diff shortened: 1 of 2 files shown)


UziTech (changes requested):
Nice catch! 💯  Thanks for doing this

Issue #704, opened 2023-03-02
Detecting pen force


milengg:
I see that there is new API that allow detect force of the pen - touch.force. Is it possible to use such thing to receive more authentic sign?

UziTech:
Pressure is already a property in the data

UziTech:
Currently only the velocity is used to determine the stroke width but we could use the force if we can figure out a good algorithm.

https://github.com/szimek/signature_pad/blob/11272b4a7ac0fe5ca88cba08b5046b8decc1201a/src/signature_pad.ts#L500-L502

BinoDino:
Hi, 
is there any update on this issue?

UziTech:
I don't think anyone has been working this.

PR #880, opened 2026-06-26
feat: add pressure detection to SignaturePad

xtruhlar:
Closes [#704](https://github.com/xtruhlar/signature_pad/issues/704).

The `pressure` field already existed on `BasicPoint` but the values being stored weren't always meaningful. Specifically, when using a regular mouse, browsers report 0.5 for `PointerEvent.pressure` whenever a button is pressed in my opinion that's not real hardware pressure, it's just a spec default. This change makes the library aware of the difference.

What changed:
1. Mouse pointer events `(pointerType: 'mouse')` now correctly report `pressure: 0` instead of the misleading 0.5 the browser sends
2. Pen and touch pointer events still read `event.pressure` as before, since those can carry real hardware data
3. Touch events still read `touch.force`, which reflects actual pressure on devices that support it (e.g. older iPhones with 3D Touch)
4. Added a `hasPressure` boolean getter that turns `true` the moment a stroke with real pressure data comes in which is useful if you want to know whether the device actually supports it
5. `hasPressure` resets to `false` on `clear()`
6. GUI enhancement - more or less for debugging:
image

Tested on:

- Desktop with mouse: `hasPressure` stays `false`, pressure correctly 0 ✓
- Stylus:  `hasPressure` becomes `true`, real pressure values 0.0–1.0 reflected in point data ✓
- I suggest more tests with different HW devices.

Tests added covering all four cases: default state, mouse strokes, pen strokes, and reset on clear().


xtruhlar:
Thank for your response @szimek, I tried to resolve all problem Copilot mentioned. Can you check it again?

UziTech:
I'm curious why you want this change? There is nothing preventing a mouse from providing the actual pressure, and there is nothing that says a stylus or pen has to provide pressure. So this doesn't seem to be accurate. So why should this be added at all?

xtruhlar:
Hey, I wanted to use this for eSinging where the pressure (based on hardware) would be great source of information to authorities if the person must be linked with the sign as detailed as possible. So the replay is great option but additional pressure information if provided would make it more inspectable.

UziTech:
Since we do provide the full event object in the listeners I think this type of calculation can be done without any changes to signature_pad. Or you can inspect the data after and check if pressure is always 0.5 (probably more accurate)

We are trying to keep signature pad to only the features it needs to be able to draw the signature. Anything else should be done outside of signature_pad with the data we provide.

Signature_pad should just be one part of a full esigning application. We are not trying to make it a stand-alone application.

PR #890, opened 2026-07-28
fix: render two-point point groups

### LINKED ISSUE #720: Point groups with only 2 points are not drawn
_If you have a question, please ask it on [StackOverflow](https://stackoverflow.com/questions/ask?tags=javascript+signature-pad) and use `signature-pad` tag instead of creating an issue._

Bug

If a point group only has 2 points, nothing gets drawn.

_If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via <https://jsfiddle.net> or similar (template: <https://jsfiddle.net/szimek/h80phxyr/>)._
https://jsfiddle.net/xvoqnyu9/16/
Not quite sure why clicking the buttons multiple times crashes signature_pad. So jus refresh (or save) before clicking each button. You can see the one with 2 points does not render anything while the one with 3 points renders a line.

Even with 2 points it should render a line.

As far as I know it hasn't worked previously. 4.1.5 affected, chrome latest.

EXPLANATION:
signature_pad.ts line 575 allows a point group with 2 points to go through to the _addPoint function (for the curve).

signature_pad.ts line 454 does not allow something with 2 points to go through to the rest of the logic. This is fine if there are more points, but if there are ONLY 2, then nothing happens and the logic ends there. The point group is essentially ignored.

I'm not sure if this is intentional. Maybe the customer we have that provided this point group just needs to provide better point groups.

11suixing11:
## Summary

- Render point groups with exactly two points as a round-capped line in `fromData()` and `redraw()`.
- Emit an SVG `<line>` for two-point groups so SVG export no longer drops them.
- Use the same point-group width rules for Canvas and SVG, including custom `dotSize`.
- Add behavior coverage for Canvas path geometry, stroke count, line width, round caps, SVG attributes, and the non-empty state.
- Rebuild the tracked docs bundle to match the source change.

Fixes #720.

## Root cause

The `fromData()` and `toSVG()` rendering path only produced Bezier curves after more than two points, while one-point groups were handled as dots. Groups with exactly two points entered the curve path but never had enough points to create a Bezier, so nothing was drawn or exported.

## Verification

- `corepack yarn test tests/signature_pad.test.ts --runInBand` (48 tests)
- `corepack yarn lint`
- `corepack yarn test --runInBand` (4 suites, 62 tests, 11 snapshots)
- `corepack yarn prettier --check src/signature_pad.ts tests/signature_pad.test.ts`
- `corepack yarn build` (tracked docs bundle reproduced with no diff)
- `git diff --check`


UziTech (approved):
Thanks for this! 💯

